PR #1, opened 2021-12-29
Add GQL codegen

sergiimk:
@kulykivska


src/app/api/search.api.ts
```diff
@@ -106,38 +107,17 @@ export class SearchApi {
...
-            }));
-    }
-    private static searchValueAddToAutocomplete(ngTypeaheadList: DatasetIDsInterface[], searchValue: string): DatasetIDsInterface[] {
-        const newArray: DatasetIDsInterface[] = JSON.parse(JSON.stringify(ngTypeaheadList));
-        if (searchValue) {
-            newArray.unshift({__typename: TypeNames.allDataType, id: searchValue});
-        }
-        return newArray;
+        return this.searchAutocompleteGQL.watch({query: id, perPage: 10}).valueChanges.pipe(map(result => {
+            let results = result.data.search.query.nodes.map(
```
kulykivska, line 112: Identifier `results` is never reassigned. Should be created new typization variable with let. For example:
```
let nodesList: DatasetIDsInterface [] = result.data.search.query.nodes.map(
v => ({id: v.name, __typename: v.__typename as TypeNames}));
if (searchValue) {
   nodesList.unshift({__typename: TypeNames.allDataType, id: id});
}
return nodesList;
```

src/app/api/search.api.ts
```diff
@@ -106,38 +107,17 @@ export class SearchApi {
...
-                }
-            }));
-    }
-    private static searchValueAddToAutocomplete(ngTypeaheadList: DatasetIDsInterface[], searchValue: string): DatasetIDsInterface[] {
-        const newArray: DatasetIDsInterface[] = JSON.parse(JSON.stringify(ngTypeaheadList));
-        if (searchValue) {
-            newArray.unshift({__typename: TypeNames.allDataType, id: searchValue});
-        }
-        return newArray;
+        return this.searchAutocompleteGQL.watch({query: id, perPage: 10}).valueChanges.pipe(map(result => {
```
kulykivska, line 111: result should be with type

src/app/api/kamu.graphql.ts
```diff
@@ -0,0 +1,376 @@
+// THIS FILE IS GENERATED, DO NOT EDIT!
+import { gql } from '@apollo/client/core';
```
kulykivska, line 2: File with interfaces should be named like: `kamu.graphql.interface.ts `

package.json
```diff
@@ -8,7 +8,9 @@
     "test": "ng test",
     "lint": "ng lint",
     "e2e": "ng e2e",
-    "postinstall": "ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points"
+    "postinstall": "ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points",
+    "gql-update-schema": "npx rover graph introspect http://localhost:8080/graphql > resources/schema.graphql",
```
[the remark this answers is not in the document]
kulykivska: What is `npx` version are you using? Could you note it in `README.md` file?

kulykivska (changes requested):

src/app/api/kamu.graphql.ts
```diff
@@ -0,0 +1,376 @@
...
+import { Injectable } from '@angular/core';
+import * as Apollo from 'apollo-angular';
+export type Maybe<T> = T | null;
+export type InputMaybe<T> = Maybe<T>;
+export type Exact<T extends { [key: string]: unknown }> = { [K in keyof T]: T[K] };
+export type MakeOptional<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]?: Maybe<T[SubKey]> };
+export type MakeMaybe<T, K extends keyof T> = Omit<T, K> & { [SubKey in K]: Maybe<T[SubKey]> };
+/** All built-in and custom scalars, mapped to their actual values */
+export type Scalars = {
```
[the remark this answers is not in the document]
sergiimk: Not sure it's possible with this codegen. The configuration of a TypeScript codegen plugin is here but I did not find anything type/interface related: https://www.graphql-code-generator.com/plugins/typescript

package.json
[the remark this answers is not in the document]
sergiimk: I added instructions to readme and the missing `apollo/rover` dev dependency.

I believe `npx` is a built-in command of `npm`. I have version `6.14.15` as came with node v12 installed via `nvm`.

src/app/api/kamu.graphql.ts
```diff
@@ -0,0 +1,376 @@
...
+
+export type SearchAutocompleteQueryVariables = Exact<{
+  query: Scalars['String'];
+  perPage?: InputMaybe<Scalars['Int']>;
+}>;
+
+
+export type SearchAutocompleteQuery = { __typename?: 'Query', search: { __typename?: 'Search', query: { __typename?: 'SearchResultConnection', nodes: Array<{ __typename: 'Dataset', id: any, name: string, kind: DatasetKind }> } } };
+
+export const SearchAutocompleteDocument = gql`
```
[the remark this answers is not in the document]
sergiimk: Codegen dumps everything into a single file. Multi-file generator seems to be [deprecated](https://github.com/dotansimha/graphql-code-generator/issues/703).

src/app/api/kamu.graphql.ts
sergiimk: Please rename file as you see fit before merging (in `gql-codegen.yml`) as I don't quite understand the naming scheme you're using.

As I mentioned above don't think I can make codegen produce interfaces so `*.interface.ts` might not be appropriate.

src/app/api/search.api.ts
sergiimk: Might be misunderstanding something, but isn't return value already typechecked as `Observable<DatasetIDsInterface[]>` as per function signature?

sergiimk: Also my intention in this PR was to show an example of using a query generated from a template. I think ideally we would remove the hand-written `DatasetIDsInterface` type and return a type generated from the query as well.

This way any time we update the query - everything that depends on it will see the updated type.

src/app/api/search.api.ts
sergiimk: I think typechecking works just fine here. The function return type `Observable<DatasetIDsInterface[]>` already requires that `map()` operation has to produce `DatasetIDsInterface[]`.

As for:
```ts
if (searchValue) {
   nodesList.unshift({__typename: TypeNames.allDataType, id: id});
}
```
I intentionally changed the behavior - when you type `foobar` the search should still offer you an option to search for `foobar` in all datasets, even if autocomplete returned empty result.

src/app/api/kamu.graphql.ts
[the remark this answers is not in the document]
kulykivska: okay

package.json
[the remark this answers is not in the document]
kulykivska: Ok thank you

src/app/api/kamu.graphql.ts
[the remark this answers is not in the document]
kulykivska: so sad. Okay

src/app/api/search.api.ts
kulykivska: Ok. I agree

PR #15, opened 2022-03-20
Kamu 60

kulykivska:


src/app/dataset-view/dataset.component.ts
```diff
@@ -55,11 +58,58 @@ export class DatasetComponent implements OnInit, AfterContentInit, OnDestroy {
     public currentSchema: DataViewSchema;
     public isMinimizeSearchAdditionalButtons = false;
     public datasetViewType: DatasetViewTypeEnum = DatasetViewTypeEnum.overview;
-    public searchAdditionalButtonsData: SearchAdditionalButtonInterface[] = [
-        {
-            textButton: searchAdditionalButtonsEnum.DeriveFrom,
-        },
-    ];
+    public searchAdditionalButtonsData: SearchAdditionalHeaderButtonInterface[] =
```
sergiimk, line 61: I think this generic nature of the "additional" buttons is not very helpful. The set of buttons doesn't change and this only makes code more complex. I think in future we should move them to a plain HTML inside of the component.

sergiimk (approved):
Again too many changes in one PR, but changes look good. Go ahead and merge :+1:

PR #24, opened 2022-08-02
Synchronized GraphQL package to the latest kamu-cli revision

zaychenko-sergei:


kulykivska (approved):

gql-codegen.yml
```diff
@@ -1,12 +1,10 @@
 #schema: "http://localhost:8080/graphql"
 schema: "resources/schema.graphql"
-documents: "src/**/*.graphql"
 overwrite: true
+documents: "src/app/api/**/*.graphql"
 generates:
   src/app/api/kamu.graphql.interface.ts:
     plugins:
-      - add:
-          content: '// THIS FILE IS GENERATED, DO NOT EDIT!'
```
sergiimk, line 9: Why are we removing this? Seems like a helpful warning to have in a file header.

sergiimk (approved):
LGTM :+1:

gql-codegen.yml
zaychenko-sergei: Restored

PR #25, opened 2022-08-03
Simpler GraphQL update procedure directly from kamu-cli

zaychenko-sergei:
Got rid of Rover, downloading definition from `kamu-cli` repository instead.
Implemented automatic adjustment of `import { gql }` generated statement to avoid manual edits.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -651,17 +651,6 @@
         "zen-observable": "^0.8.14"
       }
     },
-    "@apollo/rover": {
-      "version": "0.4.1",
-      "resolved": "https://registry.npmjs.org/@apollo/rover/-/rover-0.4.1.tgz",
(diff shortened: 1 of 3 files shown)


sergiimk (approved):
Looks good, my only concern is that this ties frontend dev environment to Linux / Mac.

As long as this works for @kulykivska - I'm OK with it. We can deal with Windows support later when needed.

zaychenko-sergei:
> Looks good, my only concern is that this ties frontend dev environment to Linux / Mac.
> 
> As long as this works for @kulykivska - I'm OK with it. We can deal with Windows support later when needed.

Yulia works on Mac. Windows developers always have Cygwin workaround. I think in the worst case, 1st command succeeds, and even if 2nd fails, that poor Windows user will have to repeat what we did manually at most.

PR #28, opened 2022-08-09
Refactoring for Dataset View Component

kulykivska:


src/app/dataset-view/dataset.component.ts
```diff
@@ -34,64 +34,14 @@ import { Dataset, DatasetKind } from "../api/kamu.graphql.interface";
 export class DatasetComponent implements OnInit, OnDestroy {
     @ViewChild("sidenav", { static: true }) public sidenav?: MatSidenav;
     @ViewChild("menuTrigger") trigger: any;
+    @ViewChild(DatasetViewMenuComponent) datasetViewMenuComponent!: any;
```
zaychenko-sergei, line 37: Unused field

src/app/dataset-view/dataset.component.ts
```diff
@@ -100,11 +50,10 @@ export class DatasetComponent implements OnInit, OnDestroy {
     public isAvailableLinageGraph = false;
     public headings: Element[] | undefined;
     public isMarkdownEditView = false;
-    public clipboardKamuCli = "kamu pull kamu.dev/anonymous/dataset";
-    public clipboardKafka = "https://api.kamu.dev/kafka/anonymous/dataset";
     public markdownText = AppValues.markdownContain;
 
     private w: Window;
+    private subs: SubSink = new SubSink();
```
zaychenko-sergei, line 56: Not all subscriptions in this component became part of the `subs`

src/app/dataset-view/dataset.component.ts
```diff
@@ -388,6 +282,53 @@ export class DatasetComponent implements OnInit, OnDestroy {
         this.onSelectDataset(idDataset);
     }
 
+    public getParentMethod(): { [key: string]: (...args: any[]) => any } {
```
zaychenko-sergei, line 285: This should become a delegate interface instead of a dictionary

src/app/dataset-view/dataset.component.ts
```diff
@@ -34,64 +34,14 @@ import { Dataset, DatasetKind } from "../api/kamu.graphql.interface";
 export class DatasetComponent implements OnInit, OnDestroy {
```
zaychenko-sergei, line 34: The following fields and related behavior repeat between many components:
```
@ViewChild("sidenav", { static: true }) public sidenav?: MatSidenav;
public isMinimizeSearchAdditionalButtons = false;
```

Also a few more fields/behaviors repeat between main window components - `SearchComponent` and `DatasetComponent`:

```
public isMobileView = false;
public searchValue = "";
private w: Window;
```
Maybe new `subs` would also belong to these categories.

I think we should stop copy pasting these, and instead create a hierarchy of base abstract components. Perhaps 2 levels:
 - very general base (shared between all components)
 - main window base (extends the one above, generalizes common parts of `DatasetComponent` and `SearchComponent`)
 
The ultimate goal is not to have duplicate peripheral functionality between components. They should concentrate on their purpose instead and share utilities in base classes and/or utility components.

kulykivska: I need to move sidenav in separate shared component. Because we use it in many places.

src/app/dataset-view/dataset.component.ts
kulykivska: fixed

src/app/dataset-view/dataset.component.ts
kulykivska: fixed

src/app/dataset-view/dataset.component.ts
kulykivska: fixed

zaychenko-sergei (approved):

PR #29, opened 2022-08-11
Kamu 89

dmitriy-borzenko:


src/app/dataset-view/dataset.component.ts
```diff
@@ -102,7 +106,22 @@ export class DatasetComponent implements OnInit, OnDestroy {
     public isMarkdownEditView = false;
     public clipboardKamuCli = "kamu pull kamu.dev/anonymous/dataset";
     public clipboardKafka = "https://api.kamu.dev/kafka/anonymous/dataset";
-    public markdownText = AppValues.markdownContain;
+    public markdownText = `## Markdown __rulez__!
```
zaychenko-sergei, line 109: Why inlined the constant ?

src/app/services/navigation.service.ts
```diff
@@ -0,0 +1,34 @@
+import { Injectable } from "@angular/core";
+import { Router } from "@angular/router";
+import AppValues from "../common/app.values";
+import ProjectLinks from "../project-links";
+
+@Injectable()
+export class NavigationService {
+    constructor(private router: Router) {}
+
+    public navigateToHome(params?: object): void {
```
zaychenko-sergei, line 10: As these parameters in all methods are associated with particular route, can we make specific arguments instead of `object` ?

src/app/services/navigation.service.ts
```diff
@@ -0,0 +1,34 @@
...
+        this.router.navigate([ProjectLinks.urlHome], { queryParams: params });
+    }
+
+    public navigateToSearch(params?: object): void {
+        this.router.navigate([ProjectLinks.urlSearch], { queryParams: params });
+    }
+    public navigateToDatasetCreate(params?: object): void {
```
zaychenko-sergei, line 17: Empty line between functions is missing in some cases in this file. Did you fully install the formatting pre-commit hook?

src/app/dataset-view/dataset.component.html
```diff
@@ -40,13 +40,8 @@
...
                             "
                             data-pjax="#repo-content-pjax-container"
-                            [href]="
-                                '/' +
-                                datasetName.owner.name +
-                                '/dataset-view?id=' +
-                                datasetName.id +
-                                '&type=overview'
-                            "
+                            [routerLink]="[datasetLink]"
```
zaychenko-sergei, line 43: Does it have to be `href` in general? Can we replace it on click event ?

zaychenko-sergei (commented):
I think there are more places with `href` with manual URL. You've only touched 1 of 5:

image

src/app/services/navigation.service.ts
dmitriy-borzenko: In several cases, these are different parameters and we do not know what will be in the future

src/app/dataset-view/dataset.component.ts
dmitriy-borzenko: Fixed, moved to constants files

src/app/services/navigation.service.ts
zaychenko-sergei: Give me an example, where same route has different parameter kinds

zaychenko-sergei (approved):

sergiimk (approved):
Few minor comments, but overall approach looks good. :+1: 

Also please write descriptive PR messages, don't leave them blank. This is a public project and not everyone will have access to our Jira.

Good to merge once the comments are resolved.

src/app/services/navigation.service.ts
dmitriy-borzenko: Added srecific arguments instead of object

src/app/components/timeline-component/timeline.component.ts
```diff
@@ -15,7 +16,14 @@ export class TimelineComponent {
     @Input() public history: MetadataBlockFragment[];
     @Input() public pageInfo: PageBasedInfo;
 
-    constructor(public dataHelpers: DataHelpersService) {}
+    constructor(
+        public dataHelpers: DataHelpersService,
+        private navigationService: NavigationService,
+    ) {}
+
+    public showOwnerPage(ownerName: string): void {
```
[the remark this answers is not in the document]
dmitriy-borzenko: Fixed

PR #40, opened 2022-08-27
Kamu 99: Eliminate improper type casts in lineage graph component

zaychenko-sergei:
- Reworked datastructures used to prepare lineage graph data
- Unified naming of lineage-related items
- Restructured lineage graph update responsibilities in similar manner as other dataset tabs
- Eliminated redundand lineage graph resets

### DIFF
--- a/src/app/dataset-view/dataset.service.ts
+++ b/src/app/dataset-view/dataset.service.ts
@@ -20,14 +20,14 @@ import {
     PageBasedInfo,
 } from "../api/kamu.graphql.interface";
 import { ModalService } from "../components/modal/modal.service";
-import { AppDatasetSubsService } from "./datasetSubs.service";
+import { AppDatasetSubscriptionsService } from "./dataset.subscriptions.service";
 import {
     DatasetHistoryUpdate,
     DataUpdate,
     MetadataSchemaUpdate,
     DataRow,
     OverviewDataUpdate,
-} from "./datasetSubs.interface";
+} from "./dataset.subscriptions.interface";
 import { isNil } from "lodash";
 import _ from "lodash";
 import { logError } from "../common/app.helpers";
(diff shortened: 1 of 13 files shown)


src/app/dataset-view/dataset.service.ts
```diff
@@ -303,66 +230,140 @@ export class AppDatasetService {
...
+                                            return {
+                                                basics: downDependency3 as DatasetBasicsFragment,
+                                                downstreamDependencies: downDependency3.metadata.currentDownstreamDependencies.map(
+                                                    (downDependency4) => {
+                                                        return {
+                                                            basics: downDependency4 as DatasetBasicsFragment,
+                                                            downstreamDependencies: downDependency4.metadata.currentDownstreamDependencies.map(
+                                                                (downDependency5) => {
+                                                                    return {
+                                                                        basics: downDependency5 as DatasetBasicsFragment,
```
sergiimk, line 266: This probably could be expressed more concisely through recursion

src/app/dataset-view/dataset.service.ts
```diff
@@ -303,66 +230,140 @@ export class AppDatasetService {
...
+                            }
+                        )
+                    };
+                }
+            )
+        };
+     }
+
+    private updatelineageGraph(origin: DatasetLineageNode) {
+        const lineageGraphEdges: DatasetBasicsFragment[][] = [];
```
sergiimk, line 333: Should the dealings with nodes and edges be a concern of Lineage component, not dataset service? `origin` node seems to contain everything that is needed to vizualize the graph and the way it happens is not a service-level concern?

sergiimk (commented):
Great improvement :+1:

src/app/dataset-view/dataset.service.ts
zaychenko-sergei: The whole point of this is that there is no way to do it in a completely type-safe manner. First of all, there are no explicit types of parts, those generated union/intersection types do not have any name. Second each level is a different data type. So, we cannot really express it recursively with how the base GraphQL query is defined.

src/app/dataset-view/dataset.service.ts
zaychenko-sergei: At least building a traditional graph data structure should be done in the service on my opinion. The service should hide the complexity of not completely adequate GraphQL structure. But the bondary can be moved as we revise it again.

PR #44, opened 2022-09-05
0 artifficial delay on HTTP queries in production mode

zaychenko-sergei:


src/environments/environment.prod.ts
```diff
@@ -1,5 +1,5 @@
 export const environment = {
     production: true,
     github_client_id: "361a3b4fda86d0234d2f", // your Client ID from GitHub
-    delay_http_request_ms: 2000,
+    delay_http_request_ms: 0,
```
sergiimk, line 4: I noticed that in our Release CI action we do:
```
- run: npm run build
- run: npm run build-prod
```
is there a reason to run both?

sergiimk (approved):

src/environments/environment.prod.ts
zaychenko-sergei: This is not about releasing. Actually, these modes are different compilation technologies with and without optimization, and both have to work on a succesfull branch or release.  `build-prod` activates so called AOT compilation, and in fact it's catching more errors in templates, as without it some errors are only detected when template is loaded.

PR #47, opened 2022-09-08
Don't display default values in URL

dmitriy-borzenko:


zaychenko-sergei (changes requested):
1. There are 2 new linting warnings, unused imports. Please pay attention to warnings too.
2. When I am using search, I switch to page 2, then switch back to page 1, and I see `page=1` parameter. Also, I see an empty query parameter on both pages.
```
http://localhost:4200/v/search?query=&page=1
```
3. Parameter `page=1` is still visible on the History tab.
```
http://localhost:4200/kamu/alberta.case-details?tab=history&page=1
```

zaychenko-sergei (approved):

PR #49, opened 2022-09-19
Add error service

dmitriy-borzenko:


src/app/common/app.values.ts
```diff
@@ -104,3 +104,15 @@ const language = 'typescript';
             });
     }
 }
+
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
```
zaychenko-sergei, line 108: Maybe just do what it suggests - place the error texts to separate class

src/app/dataset-view/dataset.service.ts
```diff
@@ -49,19 +50,24 @@ export class AppDatasetService {
...
+            (data: GetDatasetMainDataQuery) => {
+                if (data.datasets.byOwnerAndName) {
+                    this.datasetUpdate(data.datasets.byOwnerAndName);
+                    this.overviewTabDataUpdate(data);
+                    this.dataTabDataUpdate(data);
+                    this.metadataTabDataUpdate(data);
+                    this.lineageTabDataUpdate(data);
+                } else {
+                    this.errorService.processError(
+                        new Error(ErrorTexts.ERROR_DATASET_NOT_FOUND),
```
zaychenko-sergei, line 63: Please also report a name of the dataset.
I think handling this particular error (not every) should also redirect to main page, because components are left in incompletely initialized state.

src/app/services/error.service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    public processError(error: Error): void {
+        this.errorDescription = error.message;
+
+        if (error instanceof ApolloError) {
+            this.errorDescription = error.networkError
+                ? ErrorTexts.ERROR_NETWORK_DESCRIPTION
+                : ErrorTexts.ERROR_TECHNICAL_SUPPORT;
+        }
+        if (error.message === ErrorTexts.ERROR_SERVER_BAD_SQL_QUERY) {
```
zaychenko-sergei, line 23: This piece of cod ewill quickly grow with new kinds of errors. We should have some kind of dictionary instead of `if`.

zaychenko-sergei (changes requested):

src/app/services/error.service.ts
dmitriy-borzenko: Fixed

src/app/common/errors.text.ts
```diff
@@ -0,0 +1,15 @@
...
+    public static readonly ERROR_NETWORK_DESCRIPTION =
+        "Check the Internet connection.";
+    public static readonly ERROR_DATASET_NOT_FOUND = "dataset not found.";
+    public static readonly ERROR_BAD_SQL_QUERY = "Incorrect SQL query.";
+    public static readonly ERROR_SERVER_BAD_SQL_QUERY = "DataFusion error";
+    public static readonly ERROR_TECHNICAL_SUPPORT =
+        "Operation failed. If this happens again, please contact Kamu technical support.";
+}
+
+export const DictionaryErrors = [
```
sergiimk, line 12: Unused?

src/app/services/error.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    ) {}
+
+    public processError(error: Error, datasetName?: string): void {
+        const errorMapper: { [key in string]: string } = {
+            [ErrorTexts.ERROR_SERVER_BAD_SQL_QUERY]:
+                ErrorTexts.ERROR_SERVER_BAD_SQL_QUERY,
+            [ErrorTexts.ERROR_DATASET_NOT_FOUND]: `${datasetName ?? ""} ${
+                ErrorTexts.ERROR_DATASET_NOT_FOUND
+            }`,
+        };
```
sergiimk, line 26: This string-to-string mapping seems very strange to me. Don't know what's considered a best practice in JS, but in other languages I would have an error hierarchy:
```
NotFoundError: Error {...}
DatasetNotFoundError: NotFoundError {...}

InvalidInputError: Error {...}
InvalidSQLError: InvalidInputError {...}
```
This way logic will only raise typed errors, e.g.:
```
new DatasetNotFoundError(datasetName)
```
and all message formatting will be left in this service:
```
if (error instanceof DatasetNotFoundError) {
  this.errorDescription = localize(ErrorMessages.ERROR_DATASET_NOT_FOUND, error.datasetName)
}
```

I think just the fact that your function signature is `processError(error: Error, datasetName?: string):` indicates that something is very wrong here. What would it look like in the long term?

```
processError(
  error: Error, datasetName?: string, 
  userName?: string, 
  organizationName?: string, 
  sqlQuery?: string, 
  blockHash?: string, 
  pageNumber?: string, 
  ...):
```
... you get the idea.

src/app/services/error.service.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+        this.modalService
+            .error({
+                title: ErrorTexts.ERROR_TITLE_REQUEST_FAILED,
+                message: this.errorDescription,
+                yesButtonText: "Close",
+            })
+            .then(() =>
+                error.message === ErrorTexts.ERROR_DATASET_NOT_FOUND
+                    ? this.navigationService.navigateToHome()
```
sergiimk, line 44: If dataset is not found I'd expect to see a 404 page, not a modal window.

sergiimk (changes requested):

src/app/services/error.service.ts
zaychenko-sergei: Typed errors would make sense only if we plan implementing a Visitor pattern for error handling. While we don't, at least that does not look like a reasonable effort. 

But maybe a universal error structure like this:

```
class KamuError : Error {
    errorCode: string, // could be an enum, but TS enums are quire weird
    params: string[];
};
```

and then a dictionary with correspondence between possible values of `errorCode` and a reporting template string.

src/app/dataset-view/dataset.service.ts
```diff
@@ -49,19 +53,22 @@ export class AppDatasetService {
...
+        this.datasetApi.getDatasetMainData(info).subscribe(
+            (data: GetDatasetMainDataQuery) => {
+                if (data.datasets.byOwnerAndName) {
+                    this.datasetUpdate(data.datasets.byOwnerAndName);
+                    this.overviewTabDataUpdate(data);
+                    this.dataTabDataUpdate(data);
+                    this.metadataTabDataUpdate(data);
+                    this.lineageTabDataUpdate(data);
+                } else {
+                    this.errorService.processError(new DatasetNotFoundError());
```
sergiimk, line 65: I'm OK to try out this visitor / double-dispatch pattern, but my only concern that as this PR stands this pattern is not utilized.

Every error is reported as:
```
this.errorService.processError(new DatasetNotFoundError(...));
```
so we might've as well called directly:
```
this.errorSergive.onDatasetNotFound(...);
```
Visitor will only be useful when error propagates up the chain and type information is lost at the place where it's being processed. Can you think of a place where this would happen?

src/app/services/navigation.service.ts
```diff
@@ -45,4 +50,10 @@ export class NavigationService {
     public navigateToOwnerView(ownerName: string): void {
         this.router.navigate([ownerName]).catch((e) => logError(e));
     }
+
+    public navigateToPageNotFound(): void {
+        this.router
+            .navigate([ProjectLinks.urlPageNotFound])
```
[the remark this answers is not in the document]
zaychenko-sergei: Okay, added `skipLocationChange: true` argument

src/app/dataset-view/dataset.service.ts
zaychenko-sergei: I don't agree. The visitor seems very useful in a scenario where we would want to have multiple handling strategies attached, possibly overlapping by exception type. Anyway, we've started overengineering the error handling more than a week ago. Even the step of introducing error hierarchy was not well motivated, as we haven't solved any practical problem.

zaychenko-sergei (approved):

src/app/dataset-view/dataset.service.ts
zaychenko-sergei: @sergiimk The need in the visitor immediately paid off after introducing the custom global error handler. where one no longer has knowledge on the error type.

sergiimk (approved):
:rocket:

PR #50, opened 2022-09-22
Changed changeDetection strategy for all components

dmitriy-borzenko:


src/app/dataset-view/additional-components/data-component/data-component.ts
```diff
@@ -10,6 +17,7 @@ import * as monaco from "monaco-editor";
 @Component({
     selector: "app-data",
     templateUrl: "./data-component.html",
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
zaychenko-sergei, line 20: No CDR update when new data arrives

dmitriy-borzenko: Add CDR in subscription

zaychenko-sergei (approved):

PR #57, opened 2022-11-04
Kamu 110 create account and settings page

dmitriy-borzenko:


src/app/api/auth.api.ts
```diff
@@ -48,6 +52,21 @@ export class AuthApi {
         this.userChanges$.next(user);
     }
 
+    public fetchDatasetsByAccountName(
```
zaychenko-sergei, line 55: This query is not about authentication, it's about the datasets, so please move it to the datasets API part.

src/app/app-routing.module.ts
```diff
@@ -39,6 +40,47 @@ export const routes: Routes = [
         path: ProjectLinks.URL_PAGE_NOT_FOUND,
         component: PageNotFoundComponent,
     },
+    {
+        path: ProjectLinks.URL_SETTINGS,
```
zaychenko-sergei, line 44: Can the children be routed in the same style as we have in the dataset view? Via query parameters rather than custom URL parts. We direct all these to the same component anyway.

src/app/project-links.ts
```diff
@@ -9,6 +9,7 @@ export default class ProjectLinks {
     public static readonly URL_SEARCH: string = "v/search";
     public static readonly URL_DATASET_CREATE: string = "v/new-dataset";
     public static readonly URL_PAGE_NOT_FOUND: string = "v/page-not-found";
+    public static readonly URL_SETTINGS: string = "v/settings";
```
zaychenko-sergei, line 12: I assume we no longer need `URL_PROFILE`? It's reserved, but not used.

src/app/app-routing.module.ts
zaychenko-sergei: Also, these identifiers should become constants in `ProjectLinks`

src/app/app.component.ts
```diff
@@ -150,12 +151,12 @@ export class AppComponent extends BaseComponent implements OnInit {
     }
 
     public onUserProfile(): void {
-        promiseWithCatch(
-            this.modalService.warning({
-                message: AppValues.UNIMPLEMENTED_MESSAGE,
-                yesButtonText: "Ok",
-            }),
-        );
+        if (this.authApi.currentUser?.login) {
```
zaychenko-sergei, line 154: else exception ?

src/app/app.component.ts
```diff
@@ -186,12 +187,7 @@ export class AppComponent extends BaseComponent implements OnInit {
     }
 
     public onSettings(): void {
-        promiseWithCatch(
-            this.modalService.warning({
-                message: AppValues.UNIMPLEMENTED_MESSAGE,
-                yesButtonText: "Ok",
-            }),
-        );
+        this.navigationService.navigateToSettings();
```
zaychenko-sergei, line 190: Does this also require checking if there is any user?

src/app/app-routing.module.ts
zaychenko-sergei: I think this URL also requires authenticaion guard. If it is attempted when user is not logged in, we are currently displaying empty settings page. This should redirect to home instead. Please implement a guard class, which we will use for all pages that require a logged in user.

src/app/auth/account/account.component.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            ],
+        }).compileComponents();
+
+        fixture = TestBed.createComponent(AccountComponent);
+        component = fixture.componentInstance;
+        fixture.detectChanges();
+    });
+
+    it("should create", () => {
+        expect(component).toBeTruthy();
```
zaychenko-sergei, line 41: Definitely not enough tests :-)

src/app/auth/account/account.component.ts
```diff
@@ -1,8 +1,190 @@
...
+    styleUrls: ["./account.component.sass"],
     changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class AccountComponent {}
+export class AccountComponent extends BaseComponent implements OnInit {
+    public accountViewType = AccountTabs.overview;
+    private _window: Window;
+    private resizeId: NodeJS.Timer;
+    private menuRowWidth: number;
+    public user: MaybeNull<AccountDetailsFragment>;
```
zaychenko-sergei, line 44: Under what condition user can be null?

src/app/auth/account/account.component.ts
```diff
@@ -1,8 +1,190 @@
...
     selector: "app-account",
     templateUrl: "./account.component.html",
+    styleUrls: ["./account.component.sass"],
     changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class AccountComponent {}
+export class AccountComponent extends BaseComponent implements OnInit {
+    public accountViewType = AccountTabs.overview;
+    private _window: Window;
+    private resizeId: NodeJS.Timer;
```
zaychenko-sergei, line 42: `resizeId` and `menuRowWidth` are unused

src/app/auth/account/account.component.ts
```diff
@@ -1,8 +1,190 @@
...
+        promiseWithCatch(
+            this.modalService.warning({
+                message: AppValues.UNIMPLEMENTED_MESSAGE,
+                yesButtonText: "Ok",
+            }),
+        );
+    }
+
+    public selectedTabs(accountTabKey: AccountTabs): boolean {
+        if (this._window.location.search.includes(accountTabKey)) {
```
zaychenko-sergei, line 123: Base component class has `searchString` helper. You will not need `_window` property if you use it instead.

src/app/auth/account/account.component.html
```diff
@@ -1 +1,174 @@
...
+                            class="counter"
+                            >2</span
+                        >
+                    </mat-button-toggle>
+                </mat-button-toggle-group>
+            </div>
+            <ng-container *ngIf="isAccountViewTypeOverview">
+                <div>Overview</div>
+            </ng-container>
+            <ng-container *ngIf="isAccountViewTypeDatasets">
```
zaychenko-sergei, line 141: Potenttially, the functionality will grow. So, let's move the "datasets" tab into a separate child component. It's okay to keep the other not-implemented tabs as is for now, but as soon as there will be some functionality there, those should also be in the separate components. Similarly to dataset view tabs

zaychenko-sergei: The pagination control also needs to move to this new component

src/app/auth/account/account.component.ts
```diff
@@ -1,8 +1,190 @@
...
+        private authApi: AuthApi,
+        private route: ActivatedRoute,
+        private router: Router,
+        private navigationService: NavigationService,
+        private cdr: ChangeDetectorRef,
+        private modalService: ModalService,
+    ) {
+        super();
+        this._window = window;
+        this.user = this.authApi.currentUser;
```
zaychenko-sergei, line 67: I think the user should come from the URL, not from the login.  We might be viewing someone else's account, not necessarily the account of the currently logged in user.

src/app/auth/account/account.component.ts
```diff
@@ -1,8 +1,190 @@
...
+            params.currentPage,
+        );
+    }
+
+    private navigateTo(type: AccountTabs): void {
+        this.navigationService.navigateToOwnerView(this.accountName, type);
+        this.accountViewType = type;
+    }
+
+    private getDatasets(): void {
```
zaychenko-sergei, line 181: Please create a little service, components should not be interacting with GraphQL directly.

src/app/auth/settings/settings.component.html
```diff
@@ -0,0 +1,436 @@
+<div class="pt-4 container-xl p-responsive clearfix">
+    <div class="d-md-flex align-items-center justify-content-between mt-1 mb-4">
+        <div class="d-flex align-items-center mb-2 mb-md-0">
+            <img
+                [src]="user?.avatarUrl"
+                alt="@kulykivska"
```
zaychenko-sergei, line 6: Suggest another `alt`

src/app/auth/settings/settings.component.html
```diff
@@ -0,0 +1,436 @@
...
+                                <a
+                                    routerLinkActive="active"
+                                    routerLink="/v/settings/appearance"
+                                    data-view-component="true"
+                                    class="ActionList-content ActionList-content--visual16"
+                                >
+                                    <span
+                                        class="ActionList-item-visual ActionList-item-visual--leading text-muted mr-2"
+                                    >
+                                        <svg
```
zaychenko-sergei, line 142: Can SVG live in the assets part as a static file?

src/app/auth/settings/settings.component.ts
```diff
@@ -0,0 +1,40 @@
...
+import { BaseComponent } from "src/app/common/base.component";
+
+@Component({
+    selector: "app-settings",
+    templateUrl: "./settings.component.html",
+    styleUrls: ["./settings.component.sass"],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SettingsComponent extends BaseComponent implements OnInit {
+    public activeTab: string | undefined = SettingsTabs.PROFILE;
```
zaychenko-sergei, line 17: How it can be undefined, if you have a default value always?

src/app/auth/settings/settings.component.ts
```diff
@@ -0,0 +1,40 @@
...
+    public settingsTabs: typeof SettingsTabs = SettingsTabs;
+    public user: MaybeNull<AccountDetailsFragment>;
```
zaychenko-sergei, line 19: Let's ensure by design of routing that the user is never null

src/app/auth/settings/settings.constants.ts
```diff
@@ -0,0 +1,11 @@
+export enum SettingsTabs {
+    PROFILE = "profile",
+    ACCOUNT = "admin",
+    APPEARANCE = "appearance",
+    ACCESSIBILITY = "accessibility",
+    NOTIFICATIONS = "notifications",
+    BILLING = "billing",
```
zaychenko-sergei, line 7: I think there was a popup menu item for billing, can we link to here? Please review other popup elements, check whether there are more links that can be enabled.

src/app/components/dataset-list-item/dataset-list-item.component.html
```diff
@@ -0,0 +1,127 @@
...
+                (click)="isClickableRow && onSelectDataset(row)"
+            >
+                {{ row.owner.name }}/{{ row.name }}
+            </button>
+        </div>
+        <div>
+            <ng-template #popTitle class="signed-commit__popup-header">
+                <div class="signed-commit-header d-grid">
+                    <div class="pr-1">
+                        <svg
```
zaychenko-sergei, line 18: Similarly, can SVG live in a separate static file?

zaychenko-sergei:
Couple more notes:
-  "Your Datasets" popup menu now can go to the datasets tab of the account component
- This also looks like a place for the user page link:
image

src/app/auth/account/account.component.ts
```diff
@@ -1,8 +1,186 @@
...
+                }),
+            this.accountService.onDatasetsChanges.subscribe(
+                (data: DatasetsAccountResponse) => {
+                    this.datasets = data.datasets;
+                    this.pageInfo = data.pageInfo;
+                    this.datasetTotalCount = data.datasetTotalCount;
+                    this.cdr.detectChanges();
+                },
+            ),
+            // this.accountService
```
zaychenko-sergei, line 98: Please clean the commented out code you no longer need

zaychenko-sergei (changes requested):

PR #61, opened 2022-12-09
Added new component DisplayTime.Integrated new component in the appli…

dmitriy-borzenko:
…cation.

### DIFF
--- a/src/app/components/display-time/display-time.component.ts
+++ b/src/app/components/display-time/display-time.component.ts
@@ -1,0 +1,18 @@
+import { ChangeDetectionStrategy, Component, Input } from "@angular/core";
+import { DataHelpers } from "src/app/common/data.helpers";
+
+@Component({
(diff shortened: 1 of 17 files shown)


src/app/components/display-time/display-time.component.ts
```diff
@@ -0,0 +1,43 @@
...
+    get relativeTime(): string {
+        return this.convertToRelativeTime(this.value);
+    }
+
+    private dateTime(rfc3339: string): string {
+        const dt = moment(rfc3339);
+        return dt.format(AppValues.DISPLAY_DATE_FORMAT);
+    }
+
+    public convertToRelativeTime(
```
zaychenko-sergei, line 25: Needs to be public?

dmitriy-borzenko: private is better, thanks

PR #64, opened 2022-12-13
Added fallback Event Details for unsupported event types

dmitriy-borzenko:


src/app/dataset-block/metadata-block/components/event-details/supported.events.ts
```diff
@@ -0,0 +1,11 @@
+export enum SupportedEvents {
+    AddData,
```
zaychenko-sergei, line 2: Please leave only AddData for now.
We will be adding 1 by 1, once support is implemented.

dmitriy-borzenko: Fixed

zaychenko-sergei (approved):

PR #68, opened 2022-12-21
Fixed fragment DatasetTransformContent

dmitriy-borzenko:


zaychenko-sergei (changes requested):
Please add at least 1 unit test to ensure this functionality does no break in future

zaychenko-sergei (approved):

PR #83, opened 2023-02-08
Kamu 141 Safe reaction when dataset has no schema yet

zaychenko-sergei:
- Update of GraphQL schema, following CLI branch (https://github.com/kamu-data/kamu-cli/pull/123/)
- Detecting no schema passed and properly propagating to dataset viewer components
- Hiding schema-related sections, when schema is unavailable


sergiimk (approved):
lgtm from what I can tell :ship:

PR #86, opened 2023-02-16
Kamu 131 implement links between add data/execute query event details view and query tab in dataset view

dmitriy-borzenko:


src/app/dataset-view/additional-components/data-component/data-component.ts
```diff
@@ -96,5 +102,19 @@ export class DataComponent extends BaseComponent implements OnInit {
...
+
+    private buildSqlRequestCode(): void {
+        if (this.datasetBasics) {
+            this.sqlRequestCode += `'${this.datasetBasics.name as string}'`;
+            const offset = this.location.getState() as Partial<OffsetInterval>;
+            if (
+                typeof offset.start !== "undefined" &&
+                typeof offset.end !== "undefined"
+            ) {
+                this.sqlRequestCode += `\nwhere ${this.offsetColumnName}>=${offset.start} and ${this.offsetColumnName}<${offset.end}\norder by ${this.offsetColumnName} desc\nlimit 50`;
```
zaychenko-sergei, line 116: The interval is closed, so offset = end should be included.
I guess no sense in LIMIT prefix, this is more a pagination subject.

zaychenko-sergei (changes requested):

zaychenko-sergei:
```
    public static descriptionForMetadataBlock(
        block: MetadataBlockFragment,
    ): string {
        const event = block.event;
        switch (event.__typename) {
            case "AddData":
                return `Added ${
                    event.outputData.interval.end -
                    event.outputData.interval.start
                } new records`;
            case "ExecuteQuery":
                return `Transformation produced ${
                    event.queryOutputData
                        ? event.queryOutputData.interval.end -
                          event.queryOutputData.interval.start
                        : "0"
                } new records`;
```

We are incorrectly computing interval sizes in the History tab. The intervals are closed, so end should be included.

src/app/dataset-view/additional-components/data-component/data-component.ts
dmitriy-borzenko: Fixed, included.

src/app/common/data.helpers.ts
```diff
@@ -81,14 +81,16 @@ export class DataHelpers {
...
                 } new records`;
             case "ExecuteQuery":
                 return `Transformation produced ${
                     event.queryOutputData
                         ? event.queryOutputData.interval.end -
-                          event.queryOutputData.interval.start
-                        : "0"
+                          event.queryOutputData.interval.start +
+                          1
+                        : "1"
```
zaychenko-sergei, line 93: Not sure on default 1, asked @sergiimk

PR #87, opened 2023-02-20
Kamu 148 add block data from new api in yaml format

dmitriy-borzenko:


sergiimk:
Always great to see lots of code being removed! :tada:

zaychenko-sergei (approved):

PR #99, opened 2023-05-11
Kamu 160 set polling source  remove optional string properties with empty value from yaml

dmitriy-borzenko:


src/app/dataset-view/additional-components/metadata-component/components/add-polling-source/process-form.service.ts
```diff
@@ -32,4 +35,41 @@ export class ProcessFormService {
...
+            },
+        );
+        Object.entries(form[SetPollingSourceSection.FETCH]).forEach(
+            ([key, value]) => {
+                if (!value || (Array.isArray(value) && !value.length)) {
+                    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-dynamic-delete
+                    delete formGroup.value[SetPollingSourceSection.FETCH][key];
+                }
+            },
+        );
```
zaychenko-sergei, line 64: Large code duplication. Just parameterize around section key and run a loop.

zaychenko-sergei (approved):

src/app/dataset-view/additional-components/metadata-component/components/add-polling-source/process-form.service.ts
zaychenko-sergei: Resolved

PR #104, opened 2023-06-06
Add unit tests.

dmitriy-borzenko:


zaychenko-sergei (changes requested):
I'd expect more intensive coverage of these:

image

PR #107, opened 2023-07-06
Kamu 170 update all available packages

dmitriy-borzenko:


zaychenko-sergei (changes requested):
What about `devDependencies`? Mainly linter and test runners.

zaychenko-sergei (approved):

PR #112, opened 2023-07-26
Fix parse command bug and yamlTemplate state bug.

dmitriy-borzenko:


zaychenko-sergei (approved):

zaychenko-sergei (changes requested):
I assume a similar problem exists for `SetTransform` page when we edit YAML, and editing fails.

PR #127, opened 2023-08-23
Kamu UI 122 disabled run query button after error

dmitriy-borzenko:


src/app/dataset-view/additional-components/data-component/data-component.ts
```diff
@@ -51,6 +51,12 @@ export class DataComponent extends BaseComponent implements OnInit {
 
     public ngOnInit(): void {
         this.trackSubscriptions(
+            this.appDatasetSubsService.onDatasetDataSqlErrorOccured.subscribe(
+                (dataSqlErrorUpdate: DataSqlErrorUpdate) => {
+                    this.sqlErrorMarker = dataSqlErrorUpdate.error;
+                    this.cdr.markForCheck();
+                },
+            ),
```
EvilAlexei, line 59: I suggest to use pipe `async`.
I it will check for changes and unsubscribe by itself

src/assets/styles/var.sass
```diff
@@ -137,7 +137,7 @@ $app-color-action-list-item-inline-divider: #d0d7de7a
     border-radius: 1px
     position: relative
     float: left
-    border-color: $app-color-border-default
+    border-color: $app-color-border-default!important
     height: inherit
     background: $app-aliceBlue
```
EvilAlexei, line 142: Just want to mention there is an error value in the variable, there are double background and it doesn't work.
Could you also fix this?

src/assets/styles/var.sass
```diff
@@ -245,7 +245,7 @@ $app-color-action-list-item-inline-divider: #d0d7de7a
   background-color: $app-color-btn-counter-bg
 .btn:hover, .btn.hover, [open]>.btn
   background-color: $app-color-btn-hover-bg
-  border-color: $app-color-btn-hover-border
+  border-color: $app-color-btn-hover-border !important
```
EvilAlexei, line 248: Please don't use `!important`

I wasn't able to find exact place there this rule is used, but if you need help to solve this - please contact me.

EvilAlexei (changes requested):

src/assets/styles/var.sass
```diff
@@ -137,7 +137,7 @@ $app-color-action-list-item-inline-divider: #d0d7de7a
     border-radius: 1px
     position: relative
     float: left
-    border-color: $app-color-border-default
+    border-color: $app-color-border-default!important
```
EvilAlexei, line 140: Please don't use `!important`,
I assume this was for the button hover state?
Could you please just add a `:hover` rule with needed styles

EvilAlexei (changes requested):

src/app/dataset-view/additional-components/data-component/data-component.ts
zaychenko-sergei: It's a reasonable comment, however it should be done as a separate refactoring issue, and for many locations like this, as this is not the only place for a related improvement

dmitriy-borzenko: Need discuss about it.

src/assets/styles/var.sass
dmitriy-borzenko: Fixed

src/app/dataset-view/additional-components/data-component/data-component.ts
EvilAlexei: Aggree

zaychenko-sergei: Reported https://github.com/kamu-data/kamu-web-ui/issues/129

PR #128, opened 2023-08-23
`defaultScalarType` option for GraphQL code generator

zaychenko-sergei:
Enabled configuration option for GraphQL code generator - it emits 'string' instead of 'any' by default when seeing scalars.
Cleanup of related typing fallout all over the project.


sergiimk (approved):
Surprised it's not the default behavior.

Would we benefit from registering parsers for scalars like DateTime on the client side?
https://www.apollographql.com/docs/apollo-server/schema/custom-scalars/

sergiimk:
Don't mind the last link - didn't realize its for Apollo server. But I wonder if there's something similar for codegen.

PR #135, opened 2023-08-28
Kamu UI 134 updating the list of datasets after deletion

dmitriy-borzenko:


src/app/dataset-view/additional-components/settings-component/settings.component.sass
```diff
@@ -1,139 +1,140 @@
 @import src/assets/styles/var
 p
-    margin: 0
+	margin: 0
```
EvilAlexei, line 3: Could you please check `settings.component.sass`, for some reason all indents are changed from spaces to tabs, it's not consistent with the project.

EvilAlexei (changes requested):

src/app/dataset-view/additional-components/settings-component/settings.component.sass
dmitriy-borzenko: Fixed

zaychenko-sergei (approved):

PR #138, opened 2023-08-30
Feature/UI rebac

zaychenko-sergei:


src/app/api/account.api.ts
```diff
@@ -1,12 +1,28 @@
...
+                accountName,
+            })
+            .valueChanges.pipe(
+                first(),
+                map((result: ApolloQueryResult<AccountByNameQuery>) => {
+                    if (result.data.accounts.byName) {
+                        return result.data.accounts.byName;
+                    } else {
+                        return null;
+                    }
```
EvilAlexei, line 24: nit: this could be simplified
```suggestion
                    return result.data.accounts.byName || null;
```

zaychenko-sergei: Linter recommends ?? instead of ||

PR #153, opened 2023-09-08
KAMU-141: migration SASS to SCSS

EvilAlexei:
- changes sass syntax to scss
- removed empty files
- removed duplicates imports
- added global path for var.scss (and changed import)


zaychenko-sergei (approved):
Don't forget changelog next time, I've added a record

PR #154, opened 2023-09-12
KAMU-143 added stylelint

EvilAlexei:
- Added stylelint with `stylelint-config-standard-scss` lint rules for scss.
- Added new commands to the package.json: `stylelint` and `stylelint-fix`
- Added stylelint to the build and release yaml files
- Fixes for styles manual and stylelint-fix + prettier

#### Note: 
- To enable stylelint in vscode please install stylelint extension 
- In the extesion settings you need to add scss, it's not there by default (it can be done in the extension settings and in the vs code settings.json)

image
image


src/app/components/pagination-component/pagination.component.scss
```diff
@@ -39,17 +39,19 @@ ngb-pagination {
     width: 16px;
     height: 16px;
     content: '';
-    background-color: currentColor;
+    background-color: currentcolor;
```
dmitriy-borzenko, line 42: Please check, no such property exists. [https://saje.hashnode.dev/css-currentcolor](url)

src/assets/styles/var.scss
```diff
@@ -751,7 +749,7 @@ markdown {
     display: inline-block;
     overflow: visible !important;
     vertical-align: text-bottom;
-    fill: currentColor;
+    fill: currentcolor;
```
dmitriy-borzenko, line 752: The same

src/styles.scss
```diff
@@ -283,7 +310,7 @@ a {
     display: inline-block;
     overflow: visible !important;
     vertical-align: text-bottom;
-    fill: currentColor;
+    fill: currentcolor;
```
dmitriy-borzenko, line 313: The same

EvilAlexei: answered in the first comment

src/assets/styles/var.scss
EvilAlexei: answered in the first comment

src/app/components/pagination-component/pagination.component.scss
EvilAlexei: Please check other documentation
https://developer.mozilla.org/en-US/docs/Web/CSS/color#values

EvilAlexei: They both valid, but In this case camel case violates the rule of the linter for CSS properties and I see no point in adding an exception if it can be fixed

src/app/common/base-test.helpers.spec.ts
```diff
@@ -19,6 +19,7 @@ export function emitClickOnElement<T>(fixture: ComponentFixture<T>, selector: st
 
 export function emitClickOnElementByDataTestId<T>(fixture: ComponentFixture<T>, id: string): void {
     const debugElement: MaybeNull<DebugElement> = fixture.debugElement.query(By.css(`[data-test-id="${id}"]`));
+    console.log(`[data-test-id="${id}"]`);
```
dmitriy-borzenko, line 22: Is it necessary?

EvilAlexei: It's not, thank you!

Just want to mention this PR is still in Draft, I will clean it up :)

.prettierrc.json
```diff
@@ -2,21 +2,5 @@
...
-      "files": [
-        "*.html",
-        "**/*.ts"
-      ],
-      "options": {
-        "tabWidth": 4
-      }
-    }
-  ]
+  "tabWidth": 4
```
EvilAlexei, line 5: Note:
This is the same as was before.
- `*.test.ts` already have `semi: true` from the parent
- all files for prettier already have tabsWidth: 4

zaychenko-sergei (approved):
I've added an ignore file (originally it reported violations for `dist` content).
Also, I've added a line in CHANGELOG.md

PR #188, opened 2023-09-29
Feature/kamu login

zaychenko-sergei:


src/app/auth/login/login.service.ts
```diff
@@ -1,21 +1,46 @@
...
 
+    private loginCallback: (loginResponse: LoginResponse) => void = this.redirectUrlLoginCallback.bind(this);
+
+    public accessTokenObtained(): Observable<string> {
+        return this.accessTokenObtained$.asObservable();
+    }
+
+    public accountChanged(): Observable<AccountFragment> {
+        return this.accountChanged$.asObservable();
+    }
```
EvilAlexei, line 42: For meaningful naming, I suggest changing methods to properties or renaming methods

Properties:
public accessToken$ = this.accessTokenObtained$.asObservable();
public account$ = this.accountChanged$.asObservable();

or

Methods:
getAccessTokenObservable
getAccountObservable

EvilAlexei: I agree with the arguments, but I want to mention that the first option wasn't about exposing a subject,
the subject remains private, but the method changes to a property and exposes observable.

But there are no strong points about it, I saw both variants I assume it's more a matter of taste

PR #190, opened 2023-10-04
Kamu UI 139 remove view encapsulation.none and refactor styles

dmitriy-borzenko:


src/styles.scss
```diff
@@ -3,6 +3,7 @@
 @import "bootstrap-icons/font/bootstrap-icons.css";
 @import "bootstrap/scss/bootstrap";
 @import "@danielmoncada/angular-datetime-picker/assets/style/picker.min.css";
+@import "../src/assets/styles/components/index";
```
EvilAlexei, line 6: Previously I added a path alias for `var.scss`, and it should work for the index too.
```suggestion
@import "components/index";
```

src/app/dataset-view/dataset.component.scss
```diff
@@ -1,145 +1,149 @@
...
         }
 
-        .datatab-activity-timestamp {
-            color: rgb(0 0 0 / 66%);
-            flex-shrink: 0;
-            text-align: right;
-            width: 62px;
-            margin-right: 10px;
-        }
+        .datatab-expanding-row {
```
EvilAlexei, line 18: Styles that are from other components should be inside those components (if it's not a 3rd party).
In this case, it should be in the `data.component`

Please check other styles from this file.

src/app/dataset-view/dataset.component.scss
```diff
@@ -1,145 +1,149 @@
...
+            .mat-button-toggle-group-appearance-standard {
+                border-radius: 0 !important;
+            }
         }
 
-        .datatab-activity-row-summary {
-            display: flex;
+        #dataset-view-history {
+            overflow: visible;
         }
```
EvilAlexei, line 16: These styles should be in the history.component

src/app/dataset-view/dataset.component.scss
```diff
@@ -1,145 +1,149 @@
...
-        line-height: 20px;
-        cursor: pointer;
-        display: flex;
-        padding: 4px 10px;
-
-        &-content {
-            flex-grow: 1;
-            max-width: calc(100% - 24px);
+:host {
+    ::ng-deep {
```
EvilAlexei, line 4: Inside `:host ::ng-deep` should be only 3rd party styles,
styles from child components should be in their own style files

EvilAlexei (changes requested):

src/app/dataset-block/metadata-block/components/tooltip-icon/tooltip-icon.component.scss
```diff
@@ -1,26 +0,0 @@
-.tooltip-class {
```
EvilAlexei, line 1: Please remove this file
And don't forget to run stylelint along with lint  :)

I have plans to add a pre-commit hook with lint + stylelint, for safe commits :D

EvilAlexei (changes requested):

src/assets/styles/components/_tooltip.scss
```diff
@@ -0,0 +1,30 @@
...
+    margin-bottom: 5px;
+}
+
+.bs-tooltip-top,
+.bs-tooltip-bottom {
+    .tooltip-arrow {
+        &::before {
+            border-top-color: #ced4da !important;
+            border-bottom-color: #ced4da !important;
+            top: -5px !important;
```
EvilAlexei, line 23: I think it should work without `!importants` - I checked a few places, and it was OK.
But I could miss something. If there are any style collisions - please contact me, and we could try to solve them together

src/assets/styles/components/_tooltip.scss
```diff
@@ -0,0 +1,30 @@
+.tooltip-class {
+    z-index: 1000;
+}
+
+.tooltip-inner {
+    background-color: #f3f4f6;
+    position: relative;
+    font-size: 12px;
+    border: 1px solid #ced4da;
```
EvilAlexei, line 9: I suggest adding a variable for border color,
there are three places here with this color, and I think it should be the same all the time.
For example `$border-color: #ced4da;` and replace border-color with this variable in all 3 places

EvilAlexei (changes requested):

src/styles.scss
dmitriy-borzenko: Fixed

src/app/dataset-view/dataset.component.scss
dmitriy-borzenko: Fixed

src/app/dataset-block/metadata-block/components/tooltip-icon/tooltip-icon.component.scss
dmitriy-borzenko: Ok, removed

src/assets/styles/components/_tooltip.scss
dmitriy-borzenko: Yes, thanks. Removed

src/assets/styles/components/_tooltip.scss
dmitriy-borzenko: Ok. fixed

src/app/dataset-view/dataset.component.scss
dmitriy-borzenko: Moved to separate file

src/app/dataset-view/dataset.component.scss
dmitriy-borzenko: Moved to history.component

EvilAlexei (approved):

PR #204, opened 2023-10-16
Added session state for graph's side panel

dmitriy-borzenko:


src/app/services/session-storage.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from "@angular/core";
+import AppValues from "../common/app.values";
+
+@Injectable({
+    providedIn: "root",
+})
+export class SessionStorageService {
+    public get sidePanelVisibility(): string | null {
+        return sessionStorage.getItem(AppValues.SESSION_STORAGE_SIDE_PANEL_VISIBLE);
+    }
```
EvilAlexei, line 10: Why are you using a getter here?
I think just a "get" method will be enough.

```suggestion
    public getSidePanelVisibility(): string | null {
        return sessionStorage.getItem(AppValues.SESSION_STORAGE_SIDE_PANEL_VISIBLE);
    }
```

But if you have some strong points to use a getter, and it will be used as a parameter, I suggest renaming it to `isSidePanelVisible`

EvilAlexei (changes requested):

src/app/services/session-storage.service.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { TestBed } from "@angular/core/testing";
+import { SessionStorageService } from "./session-storage.service";
+
+describe("SessionStorageService", () => {
```
zaychenko-sergei, line 4: Not covering functionality. Simple, but still..

src/app/services/session-storage.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from "@angular/core";
+import AppValues from "../common/app.values";
+
+@Injectable({
+    providedIn: "root",
+})
+export class SessionStorageService {
+    public get sidePanelVisibility(): string | null {
```
zaychenko-sergei, line 8: I think you should encapsulate conversion to boolean here, and a default value when key is unavailable.
This looks bad:

```
this.showSidePanel = JSON.parse(visiblePanel) as boolean;
```

I just want to call a service hiding session storage, and obtain my boolean value as is.

dmitriy-borzenko: Moved conversion to boolean into service

src/app/services/session-storage.service.ts
dmitriy-borzenko: Fixed

src/app/services/session-storage.service.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+    it("should be created", () => {
+        expect(service).toBeTruthy();
+    });
+
+    it("should be check set state for side panel", () => {
+        service.setSidePanelVisible(true);
+        expect(service.isSidePanelVisible).toBeTrue();
+        service.setSidePanelVisible(false);
+        expect(service.isSidePanelVisible).toBeFalse();
+    });
```
zaychenko-sergei, line 21: This does not cover default value

src/app/services/session-storage.service.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from "@angular/core";
+import AppValues from "../common/app.values";
+
+@Injectable({
+    providedIn: "root",
+})
+export class SessionStorageService {
+    public get isSidePanelVisible(): boolean | null {
```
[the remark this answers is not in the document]
dmitriy-borzenko: Fixed

zaychenko-sergei (approved):

PR #212, opened 2023-10-25
Kamu UI 199 console errors on app load

dmitriy-borzenko:


src/app/app.module.ts
```diff
@@ -122,6 +119,9 @@ const Services = [
                             // For now we are faking account IDs on the server, so they are a bad caching field
                             keyFields: ["accountName"],
                         },
+                        Datasets: {
+                            merge: true,
+                        },
```
zaychenko-sergei, line 124: I don't understand the purpose of this. API never returns `Datasets` object.

src/app/app.module.ts
```diff
@@ -139,6 +139,19 @@ const Services = [
...
+                                        incoming: DatasetData,
+                                        { mergeObjects },
+                                    ) {
+                                        if (existing) {
+                                            return mergeObjects(existing, incoming);
+                                        } else {
+                                            return incoming;
+                                        }
+                                    },
+                                },
```
zaychenko-sergei, line 154: No difference with a merge function for `metadata` field, except types.
Use a generic function

zaychenko-sergei (changes requested):
When using custom SQL data tab queries:

```
dataset.component.ts:259 An error occurred! For more details, see the full error text at https://go.apollo.dev/c/err#%7B%22version%22%3A%223.8.3%22%2C%22message%22%3A12%2C%22args%22%3A%5B%5D%7D data Query either ensure all objects of type DataQueries have an ID or a custom merge function, or  Query.data {__typename: 'DataQueries', query({"dataFormat":"JSON","limit":50,"query":"select\n  *\nfrom 'kamu/account.tokens.portfolio'","queryDialect":"SQL_DATA_FUSION","schemaFormat":"PARQUET_JSON"}): {…}} {__typename: 'DataQueries', query({"dataFormat":"JSON","limit":50,"query":"select\n  offset\nfrom 'kamu/account.tokens.portfolio'","queryDialect":"SQL_DATA_FUSION","schemaFormat":"PARQUET_JSON"}): {…}}
```

PR #217, opened 2023-10-27
Kamu UI 216 schema panel on data tab should always show dataset schema

dmitriy-borzenko:


zaychenko-sergei (changes requested):
All cool, but we should have some unit test coverage for this aspect.

zaychenko-sergei (approved):

PR #225, opened 2023-11-01
Removed unnecessary usage of the importants in the style

dmitriy-borzenko:


src/app/dataset-create/dataset-create.component.scss
```diff
@@ -151,7 +151,13 @@ hr {
...
+    .switch-kind-container {
+        .description {
+            font-size: 12px;
+        }
+    }
+
     .fs-12 {
-        font-size: 12px !important;
+        font-size: 12px;
     }
```
EvilAlexei, line 162: This is a tricky one.
There is a main problem in the wrong structure of the element.
But I assume this is out of scope here to refactor a structure :)

There is already a global class fs-12, and there is no need for the same class here,
and there will be no need in the `.switch-kind-container { ... }`.

And this is a good example of where `!important` could be the right solution :)
What I mean:
 - you can add !important to the main .fs-12 class in the styles.scss
 - and remove it here inside the component, along with the `.switch-kind-container`

src/app/dataset-create/dataset-create.component.html
```diff
@@ -30,11 +30,11 @@ <h1 class="title">Create a new dataset</h1>
             </div>
         </div>
         <mat-divider></mat-divider>
-        <div class="my-4">
+        <div class="switch-kind-container my-4">
             <label class="d-flex" for="kind-root">
                 <input name="kind" type="radio" value="root" formControlName="kind" />
```
EvilAlexei, line 35: Please add the `id` same as the attribute `for` on the label
```suggestion
                <input name="kind" type="radio" value="root" id="kind-root" formControlName="kind" />
```

src/app/dataset-create/dataset-create.component.html
```diff
@@ -44,7 +44,7 @@ <h1 class="title">Create a new dataset</h1>
             <label for="kind-derivative" class="mt-2 d-flex">
                 <input name="kind" type="radio" value="derivative" formControlName="kind" />
```
EvilAlexei, line 45: Please add the `id` same as the attribute `for` on the label
```suggestion
                <input name="kind" type="radio" id="kind-derivative" value="derivative" formControlName="kind" />
```

src/app/dataset-view/additional-components/overview-component/components/edit-watermark-modal/edit-watermark-modal.component.scss
```diff
@@ -1,9 +1,9 @@
 .custom-picker {
-    z-index: 9999 !important;
+    z-index: 9999;
 }
 
 .w-65 {
-    width: 67% !important;
+    width: 67%;
```
EvilAlexei, line 6: It seems really strange :D

Class w-65 with the value 67
Could you please rename it have a garmony?

EvilAlexei (changes requested):

src/app/auth/settings/account-settings.component.html
```diff
@@ -11,7 +11,7 @@
                 class="avatar rounded-circle mr-3"
             />
             <div class="flex-auto mx-4 mt-3">
-                <div class="h3-custom lh-condensed">
+                <div>
```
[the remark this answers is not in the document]
dmitriy-borzenko: Fixed

src/app/auth/settings/account-settings.component.scss
```diff
@@ -1,5 +1,11 @@
 @import "var";
 
+ul {
```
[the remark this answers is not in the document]
dmitriy-borzenko: Fixed

src/app/dataset-create/dataset-create.component.scss
dmitriy-borzenko: Ok, fixed

src/app/dataset-create/dataset-create.component.html
dmitriy-borzenko: Added

src/app/dataset-view/additional-components/overview-component/components/edit-watermark-modal/edit-watermark-modal.component.scss
dmitriy-borzenko: Fixed

PR #241, opened 2023-11-15
Rootless nginx container image

cutwater:
This PR introduces compatibility fixes and changes to improve support of various kubernetes environments and distributions.

Related issue: https://github.com/kamu-data/helm-charts/issues/11

PR #242, opened 2023-11-15
Use unprivileged nginx base image

cutwater:


images/kamu-web-ui/rootfs/etc/nginx/conf.d/default.conf
```diff
@@ -0,0 +1,12 @@
+server {
+    listen       8080;
+    server_name  localhost;
+
+    gzip         on;
+
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html;
+        try_files $uri $uri/ /index.html;
```
sergiimk, line 10: What happened to our other configuration that used to be in `nginx.conf`?

Things like:
```
keepalive_timeout 65;
sendfile on;
```

images/kamu-web-ui/Dockerfile
```diff
@@ -15,3 +18,5 @@ RUN cd /usr/share/nginx/html && \
 
 COPY runtime-config.json /usr/share/nginx/html/assets/runtime-config.json
 COPY runtime-config-single-tenant.json /usr/share/nginx/html/assets/runtime-config-single-tenant.json
+
+USER nginx
```
sergiimk, line 22: Hm, how does this work?
- We are downloading and extracting the application files under `0 / root` user
- We don't do `chown`
- But somehow `nginx` user can read files belonging to root in the end?

images/kamu-web-ui/rootfs/etc/nginx/conf.d/default.conf
sergiimk: Looking inside the image itself looks like these parameters exist in `/etc/nginx/nginx.conf` which includes this file in `conf.d/` ... but we actually flattened the config on purpose:

It's much easier to see the entire config in git rather than try to guess what additional configuration is defined in the base image, or run into issues when base image config changes between versions.

IMO `conf.d` has no use in containerized env, as this `nginx` will always be running only one service, so flat config is much nicer.

sergiimk (changes requested):

images/kamu-web-ui/rootfs/etc/nginx/conf.d/default.conf
cutwater: In this case using separate unprivileged image gives almost zero benefit, because 95% of it still has to be replicated and maintained by us, namely the configuration file, which is exactly what has been done in #241

images/kamu-web-ui/Dockerfile
cutwater: Static data doesn't have to belong to `nginx` to be served, it is sufficient for the `nginx` user to only have read permissions to be able to serve the static files. Actually it even shouldn't be writable by the `nginx` user from a security standpoint.

It works in the same way in existing container image, where the static data is owned by the `root` user. 
When the `nginx` master process is started, it runs as `root`, but when the master process spawns worker processes, the EUID of worker processes is lowered to the `nginx` user, which is controlled by the `user nginx;` directive in the `/etc/nginx/nginx.conf` file.

cutwater:
@sergiimk Thank you for review and your comments. Considering your feedback, I can flatten the configuration or revert to #241, which one would you prefer?

sergiimk (approved):
LGTM

PR #243, opened 2023-11-22
Add 2nd tag with major and minor versions to the image in at kamu-web…

metiletan:
…-ui stage

### DIFF
--- a/images/Makefile
+++ b/images/Makefile
@@ -5,6 +5,9 @@ IMAGE_REPO = ghcr.io/kamu-data
 KAMU_WEB_UI_VERSION = $(shell cat ../package.json | jq -r '.version')
 KAMU_WEB_UI_IMAGE = $(IMAGE_REPO)/kamu-web-ui
 
+KAMU_WEB_UI_MAJOR_VERSION = $(shell echo $(KAMU_WEB_UI_VERSION) | cut -d. -f 1)
+KAMU_WEB_UI_MINOR_VERSION = $(shell echo $(KAMU_WEB_UI_VERSION) | cut -d. -f 2)
+
 #########################################################################################
 
 .PHONY: kamu-web-ui
@@ -12,6 +15,7 @@ kamu-web-ui:
 	docker build \
 		--build-arg KAMU_WEB_UI_VERSION=$(KAMU_WEB_UI_VERSION) \
 		-t $(KAMU_WEB_UI_IMAGE):$(KAMU_WEB_UI_VERSION) \
+		-t $(KAMU_WEB_UI_IMAGE):$(KAMU_WEB_UI_MAJOR_VERSION).$(KAMU_WEB_UI_MINOR_VERSION) \
 		kamu-web-ui/
 
 


images/Makefile
```diff
@@ -5,13 +5,17 @@ IMAGE_REPO = ghcr.io/kamu-data
...
+KAMU_WEB_UI_MINOR_VERSION = $(shell echo $(KAMU_WEB_UI_VERSION) | cut -d. -f 2)
+
 #########################################################################################
 
 .PHONY: kamu-web-ui
 kamu-web-ui:
 	docker build \
 		--build-arg KAMU_WEB_UI_VERSION=$(KAMU_WEB_UI_VERSION) \
 		-t $(KAMU_WEB_UI_IMAGE):$(KAMU_WEB_UI_VERSION) \
+		-t $(KAMU_WEB_UI_IMAGE):$(KAMU_WEB_UI_MAJOR_VERSION).$(KAMU_WEB_UI_MINOR_VERSION) \
```
romcheg, line 18: It's also necessary to push this new tag.

romcheg (changes requested):

images/Makefile
```diff
@@ -5,13 +5,17 @@ IMAGE_REPO = ghcr.io/kamu-data
 KAMU_WEB_UI_VERSION = $(shell cat ../package.json | jq -r '.version')
 KAMU_WEB_UI_IMAGE = $(IMAGE_REPO)/kamu-web-ui
 
+KAMU_WEB_UI_MAJOR_VERSION = $(shell echo $(KAMU_WEB_UI_VERSION) | cut -d. -f 1)
+KAMU_WEB_UI_MINOR_VERSION = $(shell echo $(KAMU_WEB_UI_VERSION) | cut -d. -f 2)
```
sergiimk, line 9: What is the overall intent here? Are you suggesting to only specify `major.minor` app version in charts, i.e. allowing the same chart to pull a newer patched application version?

Would this necessitate `pull: Always` in all deployments?

// Please include more context in PR description in future.

sergiimk (changes requested):

images/Makefile
romcheg: Note taken on the more description.
<details>
<summary>

## TL;DR

The intent is to allow version skew between charts and applications. Setting `Always` pull policy will be required, but it's not a problem. Details bellow.
</summary>
The intention is to make it possible to have a version skew between charts and applications and to avoid bumping chart versions every time there are new releases of applications and vice versa.

The charts will only reference the `major.minor` tag, therefore every x.y.* version of the chart will be compatible with x.y.* version of the app. If users want to Always have the latest patch version of applications it would indeed require them to set the image pull policy to `Always`. This imposes the necessity to keep compatibility between pods running different patch versions. The latter, however, is the basic requirement of smooth updates and rollbacks in k8s.

It's important to note that for any production environment, it is recommended, to set the image by its digest, not by the tag. Kamu charts allow that. In this case nothing but users control the strictly selected version of the image. In this scenario, the pull policy may be specified as `IfNotPresent` since digests always change.
</details>

images/Makefile
```diff
@@ -5,19 +5,24 @@ IMAGE_REPO = ghcr.io/kamu-data
...
 		-t $(KAMU_WEB_UI_IMAGE):$(KAMU_WEB_UI_VERSION) \
+		-t $(KAMU_WEB_UI_IMAGE):$(KAMU_WEB_UI_MAJOR_VERSION).$(KAMU_WEB_UI_MINOR_VERSION) \
 		kamu-web-ui/
 
 
 .PHONY: kamu-web-ui-push
 kamu-web-ui-push:
 	docker push $(KAMU_WEB_UI_IMAGE):$(KAMU_WEB_UI_VERSION)
+	docker push $(KAMU_WEB_UI_IMAGE):$(KAMU_WEB_UI_MAJOR_VERSION).$(KAMU_WEB_UI_MINOR_VERSION)
 	docker tag $(KAMU_WEB_UI_IMAGE):$(KAMU_WEB_UI_VERSION) $(KAMU_WEB_UI_IMAGE):latest
```
romcheg, line 26: @sergiimk Is there any particular reason this `latest` tag is added here and not during the build stage?

sergiimk: No reason. Ideally I wanted to avoid producing a bunch of tags locally and only tag during push, but unfortunately docker does not allow something like `docker push <image> -t <name1> -t <name2> -t <name3>`.

Feel free to move to build phase for consistency.

images/Makefile
sergiimk: > It's important to note that for any production environment, it is recommended, to set the image by its digest

This is exactly what I was wondering - whether this change was implying that we should deploy to prod while allowing version skew.

I agree this can be useful for quick setup and less important environments, but in prod I would like to know exactly which image is used by every single pod.

sergiimk (approved):

PR #244, opened 2023-11-24
[KAMU-239]: added scheduling settings

### LINKED ISSUE #239: Dataset Scheduling Settings
Use Figma [wireframe](https://www.figma.com/file/CLFcgKqBeU6Q92WvxULdIF/Platform-Frontend?type=design&node-id=1-261&mode=design&t=YS07KGO8vNJWIQ29-0).

This ticket assumes full feature implementation, including
  - creating HTML markup
  - Angular component and services
  - integrating with provided GraphQL [APIs](https://github.com/kamu-data/kamu-cli/issues/280)
  - adding unit tests at corresponding layers of architecture

The access to this feature should be disabled in case `SetPollingSource` or `SetTransform` events are not yet defined (in future, we will also check for `AddPushSource` alternative).

Notes:
1. Use feature flag `enableScheduling` in `runtime_config.json`: allow changing scheduling rules for datasets:
    - Enabled by default
    - Disabled in kamu ui mode - in this case, make the settings tab for schedules unaccessible.

2. Scheduling rules  form in the dataset “Settings” tab (new tab, below General)
    - Enable/Disable updates
    - For a root dataset with polling source - allow entering scheduling rule for INGEST flow in 2 ways:
          - time delta (update every N minutes, hours, days, weeks) 
          - CRON expression
    - For a derived dataset with a defined transformation - allow entering batching condition for EXECUTE_QUERY flow:
        - a setting to minimally await for N minutes after one of the inputs was updated
        - a setting to minimally await until N input records accumulate in each input
    - In near future we will also support root datasets with push sources, and allow entering a similar batching condition, however, this is not yet supported by the GraphQL API on the backend.
        
Setting an update schedule should be allowed only for dataset users with `canSchedule` privilege. This permission should be taken into account for the Settings tab in general, similarly to currently implemented privileges.

For CRON editing, there are some ready open-source Angular widgets, i.e.,  https://www.npmjs.com/package/ngx-cron-editor, however, it seems they are an overkill, and instead:
  - allow entering a corect CRON expression as text
  - provide validation error messages
  - display a hint when the entered CRON expression will trigger next time
  
There are some CRON utility libraries with smaller scope, i.e. https://www.npmjs.com/package/cron-parser, which might serve as good helpers for this implementation.

Note: For now, we will ignore other flow types, except INGEST and EXECUTE_QUERY, but there will be more flow types in similar settings tabs in future. New flows will likely include similar scheduling controls, so make sure these components are easy to reuse in future.

EvilAlexei:
Closes #239


zaychenko-sergei:
Obsolete

PR #256, opened 2024-01-19
Added access token to local storage

dmitriy-borzenko:


sergiimk:
Thanks for such a quick fix! :+1:

zaychenko-sergei (approved):

PR #305, opened 2024-04-16
Add RisingWave engine

sergiimk:


src/app/common/data.helpers.spec.ts
```diff
@@ -336,14 +336,21 @@ it("should check description for ExecuteTransform block (no data)", () => {
 [
     { engine: "spark", label: "Apache Spark" },
     { engine: "flink", label: "Apache Flink" },
-    { engine: "datafusion", label: "DataFusion" },
-    { engine: "unknown", label: "Unknown engine" },
```
sergiimk, line 340: @dmitriy-borzenko Note that we should always expect that the server is a few versions ahead of the UI and has more metadata events, engines etc. Currently UI was showing `risingwave` engine as `Unknown engine`, which was hiding the useful actual value in metadata from the user.

src/app/dataset-block/metadata-block/components/event-details/components/common/engine-property/engine-property.component.ts
```diff
@@ -12,7 +11,16 @@ import { BasePropertyComponent } from "../base-property/base-property.component"
...
-    public get descriptionEngine(): EventPropertyLogo {
-        return DataHelpers.descriptionForEngine(this.data);
+    public get engineLogo(): string | undefined {
+        return DataHelpers.descriptionForEngine(this.data).url_logo;
+    }
+
+    public get engineDescription(): string {
+        const engineDesc = DataHelpers.descriptionForEngine(this.data);
+        if (engineDesc.label && engineDesc.label.toLowerCase() != engineDesc.name.toLowerCase()) {
+            return `${engineDesc.label} (${engineDesc.name})`;
```
sergiimk, line 21: Idea here is to show full name, but also how exactly the name looks in metadata:
image

src/app/common/data.helpers.spec.ts
sergiimk: In other words - for things we don't support we should be trying to propagate raw information to the user.

This will especially be important when we allow custom metadata extensions where other companies can define their own events etc.

PR #316, opened 2024-05-17
Kamu UI 313 mqtt polling source support

dmitriy-borzenko:
Result:

image


zaychenko-sergei (commented):
No notes, but I don't understand the order of merge, as both branches overlap

zaychenko-sergei (approved):

PR #324, opened 2024-06-12
Kamu UI 307 implementation of ingest via file upload from UI

dmitriy-borzenko:
Result:

Image

Image

https://github.com/kamu-data/kamu-web-ui/assets/51016717/73cfb7b0-19fa-4e72-86b1-6f4f473da36d

Image

Image


src/app/common/app.types.ts
```diff
@@ -1,3 +1,22 @@
...
+    fields: [string, string][];
+}
+
+export interface UploadPerareData {
+    uploadPrepareResponse: UploadPrepareResponse;
+    bodyObject: FormData | File;
+    uploadHeaders: HttpHeaders;
+}
+
+export type UploadAvailableMethod = "POST" | "PUT";
```
zaychenko-sergei, line 22: It deserves a separate file

src/app/dataset-view/additional-components/overview-component/overview.component.scss
```diff
@@ -5,3 +5,56 @@
...
+    // .file-uploader {
+    //     opacity: 0;
+    //     padding: 4.5rem;
+    //     width: inherit;
+    //     height: 100%;
+
+    //     &:hover {
+    //         cursor: pointer;
+    //     }
+    // }
```
zaychenko-sergei, line 38: Garbage

src/app/services/file-upload.service.ts
```diff
@@ -0,0 +1,168 @@
...
+import { NavigationService } from "./navigation.service";
+
+@Injectable({
+    providedIn: "root",
+})
+export class FileUploadService {
+    private authHeaders: MaybeUndefined<{
+        [header: string]: string | string[];
+    }>;
+    private uploadToken: string;
```
zaychenko-sergei, line 22: The headers are defined in constructor, but not the token

src/app/services/file-upload.service.ts
```diff
@@ -0,0 +1,168 @@
...
+        const cache = this.injector.get(APOLLO_OPTIONS).cache;
+        if (cache) {
+            const datasetKeyFragment = DatasetApi.generateDatasetKeyFragment(
+                cache.identify(DatasetApi.generateAccountKeyFragment(datasetBasics.owner.id)),
+                datasetBasics.id,
+            );
+            cache.evict({
+                id: cache.identify(datasetKeyFragment),
+                fieldName: "data",
+            });
```
zaychenko-sergei, line 90: I think "metadata" section should be reset too.

zaychenko-sergei (changes requested):
Things to add:
1) Error handling. Upload prepare, actual upload and ingest - all these phases may fail. Currently erors are not even displayed in the Toastr.


2) Displaying busy indicator, perhaps modal, but decide on your own - we allow uploading large files (50mb) and ingestion of those takes a while too.


3) Control upload file limit. Let's take the value from runtime configuration too, have 50mb by default. `(done)`

4) When `SetPollingSource` is defined, we should hide the block with drag & drop. The same for derivative datasets. `(done)`

5) Not enough tests

src/app/common/app.types.ts
dmitriy-borzenko: Moved to a separate file

src/app/dataset-view/additional-components/overview-component/overview.component.scss
dmitriy-borzenko: Removed

src/app/services/file-upload.service.ts
dmitriy-borzenko: Removed 2 variables from service

src/app/services/file-upload.service.ts
dmitriy-borzenko: Replace on `metadata`

zaychenko-sergei (approved):

PR #326, opened 2024-06-18
Kamu UI 323 support etheriumlogs as a source kind in UI

dmitriy-borzenko:
Result:

image

image


zaychenko-sergei (changes requested):
We are missing a validation for `chainId` property - it must be an unsigned number:

```
"chainId": {
          "type": "integer",
          "format": "uint64",
          "description": "Identifier of the chain to scan logs from. This parameter may be used for RPC endpoint lookup as well as asserting that provided `nodeUrl` corresponds to the expected chain."
        },
```

zaychenko-sergei (approved):

PR #328, opened 2024-06-24
Kamu UI 327 fixing cache problem for id account

dmitriy-borzenko:


src/app/services/local-storage.service.ts
```diff
@@ -11,6 +11,10 @@ export class LocalStorageService {
         return localStorage.getItem(AppValues.LOCAL_STORAGE_LOGIN_CALLBACK_URL);
     }
 
+    public get accountId(): string | null {
+        return localStorage.getItem(AppValues.LOCAL_STORAGE_ACCOUNT_ID);
+    }
```
zaychenko-sergei, line 16: Make a property or method that does not deal with `null`. Throw an exception, if account is undefined.
It would look much more natural during operations, as you won't need dummy `""` default value, 
also we will have a stricter check for the actions that do not allow anonymous users.

src/app/dataset-view/additional-components/overview-component/services/dataset-commit.service.ts
```diff
@@ -122,13 +132,14 @@ export class DatasetCommitService {
         datasetId: string;
         watermark: string;
         datasetInfo: DatasetInfo;
+        accountId: string;
```
zaychenko-sergei, line 135: Let's preserve a logical order of arguments, in particular, put `accountId` before `datasetId` everywhere.

src/app/auth/login/login.service.ts
```diff
@@ -125,6 +125,7 @@ export class LoginService {
     }
 
     private redirectUrlLoginCallback(loginResponse: LoginResponse): void {
+        this.localStorageService.setAccountId(loginResponse.account.id);
```
zaychenko-sergei, line 128: We already have `account$` event, and `LoggedUserService` is a dedicated service for managing user data. Can't we encapsulate it there?

src/app/common/base.component.ts
```diff
@@ -5,9 +5,11 @@ import ProjectLinks from "../project-links";
 import { requireValue } from "./app.helpers";
 import { UnsubscribeOnDestroyAdapter } from "./unsubscribe.ondestroy.adapter";
 import { Observable, map } from "rxjs";
+import { LocalStorageService } from "../services/local-storage.service";
 
 export abstract class BaseComponent extends UnsubscribeOnDestroyAdapter {
     protected activatedRoute = inject(ActivatedRoute);
+    protected localStorageService = inject(LocalStorageService);
```
zaychenko-sergei, line 12: `LocalStorageService` is a very low-level utility to be used that widely. Consider binding to `LoggedUserService` instead to ask the same question.

zaychenko-sergei (changes requested):
I think you can already obtain this data from `LoggedInUserService`. 
It keeps this object and refreshes it properly already:

```
private loggedInUser: MaybeNull<AccountFragment> = null;
```

src/app/services/local-storage.service.ts
dmitriy-borzenko: Fixed, added new exception

src/app/dataset-view/additional-components/overview-component/services/dataset-commit.service.ts
dmitriy-borzenko: Fixed

src/app/dataset-view/additional-components/overview-component/services/dataset-commit.service.ts
```diff
@@ -95,10 +100,15 @@ export class DatasetCommitService {
         }
     }
 
-    public updateReadme(accountName: string, datasetName: string, content: string): Observable<void> {
+    public updateReadme(
+        accountName: string,
+        datasetName: string,
+        content: string,
+        accountId: string,
```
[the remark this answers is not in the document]
dmitriy-borzenko: Fixed

PR #330, opened 2024-06-25
Kamu UI 329 show add data block instead empty result

dmitriy-borzenko:


zaychenko-sergei (approved):

sergiimk:
Please provide descriptions on PRs and (especially) issues

PR #344, opened 2024-07-05
Kamu UI 342 metadata tab should show full source info

### LINKED ISSUE #342: Metadata tab should show full source info
The goal of "Metadata" tab is to be a summary of the current state of the dataset:
- what it's current schema is
- what are its current polling/push sources

If you look at [this dataset](https://platform.demo.kamu.dev/kamu/net.rocketpool.reth.tokens-burned/block/f1620b2c1eacc8c0eaa7eea37c8502dc71a81c9ae2fa1175951f80b562d1a96b5274a) - it has a polling source `EthereumLogs`, which is nicely displayed in all detail when you look at the individual block:
image

But if you switch to the "Metadata" tab, some of the polling source's information is missing (the `fetch` section):
image

We should update the "Metadata" tab to show full information about the current polling source or transform, just like how it would be shown in the "History" tab for the corresponding block.

dmitriy-borzenko:
Closes #342

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added access token feature
 ## Fixed
 - Fixed bug with cache(Data tab) on the deployment version
+- Extended full source info for EthereumLogs(Metadata tab)
 
 ## [0.22.0] -2024-06-27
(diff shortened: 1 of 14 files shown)


dmitriy-borzenko:
Result:

image

zaychenko-sergei (approved):

PR #351, opened 2024-07-17
Improved error message

dmitriy-borzenko:
Result:

image

### DIFF
--- a/src/app/services/file-upload.service.ts
+++ b/src/app/services/file-upload.service.ts
@@ -1,5 +1,5 @@
 import { AppConfigService } from "src/app/app-config.service";
-import { HttpClient, HttpHeaders } from "@angular/common/http";
+import { HttpClient, HttpErrorResponse, HttpHeaders } from "@angular/common/http";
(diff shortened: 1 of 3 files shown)


src/app/services/file-upload.service.ts
```diff
@@ -62,7 +62,11 @@ export class FileUploadService {
                 );
             }),
             catchError(() => {
-                throw new DatasetOperationError([new Error("File could not be loaded")]);
+                throw new FileUploadError([
+                    new Error(
+                        "File could not be loaded. Supported file types: CSV, JSON, Newline-delimited JSON, Geo JSON, Newline-delimited Geo JSON, ESRI Shapefile, Parquet",
+                    ),
+                ]);
```
zaychenko-sergei, line 69: I do not see a check for HTTP response status 415

zaychenko-sergei (approved):

src/app/services/file-upload.service.ts
dmitriy-borzenko: fixed

PR #368, opened 2024-07-26
Kamu UI 294 support open in new tabwindow shortcuts

### LINKED ISSUE #294: Support "open in new tab/window" shortcuts
When CTRL+Clicking on links (like dataset names in the search view) or on tabs in dataset view it is expected that destination URL will be opened in a new browser tab.

Similarly when I right-click on a tab e.g. in GitHub I see options like "Open link in new window", but I don't get such options in our UI.

dmitriy-borzenko:
Closes #294


zaychenko-sergei (changes requested):
1) Weird looking styles for not yet visited links:
image

2) Profile tabs should be made similar too:
image

zaychenko-sergei (approved):

PR #389, opened 2024-08-15
Kamu UI 370 create a new block reset dataset in the general tab in the dataset settings

### LINKED ISSUE #370: Create a new block "Reset dataset" in the "General" tab in the dataset settings
Create a block similar to the block for deleting a dataset.

 1. Correction of information in the tooltip for the recursive flag.
 2. In the new block 2 checkboxes are created. Parent - "Flatten metadata" (Metadata only mode) and child checkbox - "Recursive"

image

dmitriy-borzenko:
Closes #370 

Result:

image


zaychenko-sergei (commented):
I don't understand if the reset view should be working for derived datasets. Logically, why not. Right now it's displayed, but if "Metadata Only" method is chosen, we have an error from BE:

> Expected a Root dataset, but a Derivative dataset was provided

zaychenko-sergei (approved):

PR #396, opened 2024-08-22
Kamu UI 390 encoding of query parameters requests for file upload service

dmitriy-borzenko:


zaychenko-sergei (changes requested):
You need to test this.

src/app/services/file-upload.service.ts
```diff
@@ -108,9 +108,10 @@ export class FileUploadService {
     }
 
     public uploadFilePrepare(file: File): Observable<UploadPrepareResponse> {
+        const codec = new HttpUrlEncodingCodec();
         return this.http.post<UploadPrepareResponse>(
             `${this.appConfigService.apiServerHttpUrl}/platform/file/upload/prepare?fileName=` +
-                file.name +
+                `${codec.encodeValue(file.name)}` +
```
sergiimk, line 114: I think it would be much safer to use URL builder pattern:

```js
const url = new URL("https://www.example.dev/");

url.searchParams.append("fileName", file.name);
url.searchParams.append("contentLength", file.size);
url.searchParams.append("contentType", file.type);

console.log(url.href);
```

dmitriy-borzenko: Thanks

PR #438, opened 2024-09-23
added prototype for view service

dmitriy-borzenko:


src/app/auth/settings/account-settings.component.ts
```diff
@@ -23,12 +25,15 @@ export class AccountSettingsComponent extends BaseComponent implements OnInit {
 
     public activeTab: AccountSettingsTabs = AccountSettingsTabs.PROFILE;
     public user$: Observable<MaybeNull<AccountFragment>>;
+    public adminPrivelegesOn: boolean = false;
```
zaychenko-sergei, line 28: Don't manage state in a component. Always in a service

zaychenko-sergei: Perhaps `LoggedUserService` should encapsulate this state.
Do not work with low level `LocalStorageService` here.

src/app/services/elements-view.service.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public executeAction(action: EnumViewActions, datasetPermissions: DatasetPermissionsFragment | null): boolean {
+        if (action === EnumViewActions.SHOW_FLOWS_TAB_ACTION && datasetPermissions) {
+            return this.isAdminPrivelegesOn || this.datasetPermissionsService.shouldAllowFlowsTab(datasetPermissions);
+        }
+        if (action === EnumViewActions.SHOW_SETTINGS_TAB_ACTION && datasetPermissions) {
+            return (
+                this.isAdminPrivelegesOn || this.datasetPermissionsService.shouldAllowSettingsTab(datasetPermissions)
+            );
```
zaychenko-sergei, line 26: This isn't an action, it's a query

dmitriy-borzenko:
All comments processed, tried new solution

src/app/auth/logged-user.service.ts
```diff
@@ -20,6 +20,7 @@ import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
 export class LoggedUserService extends UnsubscribeDestroyRefAdapter {
     private loggedInUser: MaybeNull<AccountFragment> = null;
     private loggedInUser$: Subject<MaybeNull<AccountFragment>> = new ReplaySubject<MaybeNull<AccountFragment>>(1);
+    private adminPriveleges$: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
```
zaychenko-sergei, line 23: Correct spelling is "privilege", correct everywhere.

src/app/auth/settings/account-settings.component.ts
```diff
@@ -23,10 +25,17 @@ export class AccountSettingsComponent extends BaseComponent implements OnInit {
...
 
     private router = inject(Router);
     private route = inject(ActivatedRoute);
     private loggedUserService = inject(LoggedUserService);
+    private localStorageService = inject(LocalStorageService);
+
+    public componentData$: Observable<{
+        user: MaybeNull<AccountFragment>;
+        adminPriveleges: boolean;
+    }>;
```
zaychenko-sergei, line 38: This name tells nothing. 
Also, there cannot be privileges if there is no user.
I suggest a struct which contains these fields, and it may be null as a whole.

src/app/services/elements-view.service.ts
```diff
@@ -0,0 +1,43 @@
...
+            }),
+        );
+    }
+}
+
+export enum ViewModeElement {
+    NOT_AVAILABLE_MODE = "notAvailableMode",
+    PERMISSIONS_MODE = "permissionsMode",
+    ADMIN_MODE = "adminMode",
+}
```
zaychenko-sergei, line 43: `ViewModeElement` => `ElementAvailabilityMode`

```
export enum ElementVisibilityMode {
   UNAVAILABLE,
   AVAILABLE_VIA_PERMISSIONS,
   AVAILABLE_VIA_ADMIN_PRIVILEGES,
}
```

src/app/services/elements-view.service.ts
```diff
@@ -0,0 +1,43 @@
...
+                const permissions = datasetPermissions.permissions;
+                if (
+                    permissions.canCommit ||
+                    permissions.canDelete ||
+                    permissions.canRename ||
+                    permissions.canSchedule
+                ) {
+                    return ViewModeElement.PERMISSIONS_MODE;
+                }
+
```
zaychenko-sergei, line 32: These checks will not be the same for every element, we have to somehow declare what the element requires.

src/app/auth/logged-user.service.ts
dmitriy-borzenko: Fixed

src/app/auth/settings/account-settings.component.ts
dmitriy-borzenko: Renamed and fixed

src/app/services/elements-view.service.ts
dmitriy-borzenko: Fixed

src/app/services/elements-view.service.ts
dmitriy-borzenko: Fixed

Issue #424, opened 2024-09-12
Prettify "History" timeline


sergiimk:
Currently events timeline in "History" tab looks a bit broken:
![image](https://github.com/user-attachments/assets/d33ea38f-fc9b-41e8-a69d-cd14f5a1989b)

Applying CSS `rotate: 90deg` already makes a big difference:
![image](https://github.com/user-attachments/assets/65bdce34-504c-46c6-a7f5-10c2ffa1ee7f)

Let's prettify the "nodes" a bit to make them look as nice as flow events:
![image](https://github.com/user-attachments/assets/adee3b4e-ce74-47db-9d6e-0f3360e380af)

PR #449, opened 2024-10-02
Kamu UI 424 prettify history timeline

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/424

Result:
image


zaychenko-sergei (approved):

sergiimk:
The screenshot doesn't look very good imo. 

As I mention in the original ticket, this visual is likely an artifact rather than "artistic design":
image

Issue #423, opened 2024-09-12
Query explainer UI


sergiimk:
# Context
When querying Kamu user can ask for response to include a cryptographic commitment that will hold Kamu node that provided it forever accountable for the returned data and allow other nodes on the network to reproduce and verify the request.

See for details:
- [Verifiable Queries](https://docs.kamu.dev/node/commitments/)
- [Query REST API](https://docs.kamu.dev/node/protocols/rest-api/)

# Requirements
Client applications (LLM chat bots) that provide end-users some information based on queries executed in Kamu should be able to provide end user a link that takes them to Kamu and shows them:
- The query that was executed
- Input datasets that were used in the query
- The identity of the node that originally executed the query
- Whether the commitment is valid
- If commitment is NOT valid it should indicate what part of the commitment failed the validation (e.g. invalid signature, data hash mismatch, etc.)
- Sample of the output data
- "Verify query result"  button available only for admin

## Data Flow
- Client application queries Kamu with `include=proof` parameter and displays data to the end user
- Client application also presents end user a URL that will take them to Kamu and show the explanation and verification info of the query
- This link should somehow include the commitment data (e.g. encoded as a query parameter)

### Example
> This is just an rough idea - exact implementation TBD:

Client app makes a query:
```
echo '{
  "query": "select * from \"kamu/covid19.quebec.case-details\" limit 1",
  "include": ["proof"]
}' | xh POST "https://api.demo.kamu.dev/query"
```

Response:
```json
{
    "input": {
        "query": "select * from \"kamu/covid19.quebec.case-details\" limit 1",
        "queryDialect": "SqlDataFusion",
        "dataFormat": "JsonAoS",
        "include": [
            "Input",
            "Proof"
        ],
        "datasets": [
            {
                "id": "did:odf:fed0122c8ddb48715dbaac81f90d9d19f306cc744f69f263bd0b0b8375c1c7bdf3f0c",
                "alias": "kamu/covid19.quebec.case-details",
                "blockHash": "f162016c72ca766d9ecb5e9abeca1486f311eb127ed575dd7c7f71cc700339cf746c7"
            }
        ],
        "skip": 0,
        "limit": 100
    },
    "output": {
        "data": [
            {
                "age_group": "20-29",
                "case_status": "Recovered",
                "date_reported": "2021-04-07T12:00:00Z",
                "exposure": "Close Contact",
                "gender": "Female",
                "health_region": "Ottawa Public Health",
[...]

PR #452, opened 2024-10-17
Kamu UI 423 query explainer UI

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/423

Result:
image

image


src/app/app-routing.module.ts
```diff
@@ -46,6 +47,11 @@ export const routes: Routes = [
         path: ProjectLinks.URL_ADMIN_DASHBOARD,
         component: AdminDashboardComponent,
     },
+    {
+        canActivate: [AdminGuard],
```
zaychenko-sergei, line 51: The ticket is not saying anything that this new page should be limited to admins

src/app/components/query-explainer/query-explainer.component.html
```diff
@@ -0,0 +1,243 @@
+<ng-container *ngIf="verifyResponse">
+    <div
+        class="alert alert-info message-success mx-4 mt-4 fs-18 w-50 mx-auto d-flex align-items-center justify-content-center"
+        role="alert"
+        *ngIf="verifyResponse?.ok"
```
zaychenko-sergei, line 5: You've checked `verifyResponse` in the upper tag. If you are here, it cannot be undefined or null.
Correct everywhere.

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+export class QueryExplainerComponent extends BaseComponent implements OnInit {
+    private clipboard = inject(Clipboard);
+    private queryExplainerService = inject(QueryExplainerService);
+    private cdr = inject(ChangeDetectorRef);
+    private blockService = inject(BlockService);
+
+    private sqlQuery: MaybeNull<string>;
+    public readonly DATE_FORMAT = AppValues.DISPLAY_FLOW_DATE_FORMAT;
+
+    public sqlQueryExplainer: QueryExplainerResponse;
```
zaychenko-sergei, line 31: Use word "response" in the variable name too. Right now it looks like it's a component

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+    public sqlQueryVerify$: Observable<VerifyQueryResponse>;
+    public verifyResponse: VerifyQueryResponse;
```
zaychenko-sergei, line 33: So input `sqlQuery` may be null, but the resulting data can't? I'm not sure this typing is accurate.

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+    ngOnInit(): void {
+        this.sqlQuery = this.extractSqlQueryFromRoute();
+
+        if (this.sqlQuery) {
+            const blockHashObservables$: Observable<string>[] = [];
+            this.queryExplainerService
+                .proccessQuery(this.sqlQuery)
+                .pipe(
+                    tap((response) => {
+                        this.sqlQueryExplainer = response;
```
zaychenko-sergei, line 45: Why can't we move the response into a separate observable, perhaps using `sharedReplay` operator to reuse the value?

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+                            });
+                    }),
+                    switchMap((response: QueryExplainerResponse) => {
+                        return this.queryExplainerService.verifyQuery(response);
+                    }),
+                    catchError((e: HttpErrorResponse) => {
+                        return of(e.error as VerifyQueryResponse);
+                    }),
+                    tap((data) => {
+                        this.verifyResponse = data;
```
zaychenko-sergei, line 61: Again, this should be stored as an observable.

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+                    }),
+                    switchMap(() => forkJoin(blockHashObservables$)),
```
zaychenko-sergei, line 63: `blockHashObservables$` as a local variable... this is whatever, but not functional reactive programming.

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+            const currentElementChildren: HTMLCollectionOf<HTMLElement> =
+                currentElement.children as HTMLCollectionOf<HTMLElement>;
+            setTimeout(() => {
+                currentElementChildren[0].style.display = "inline-block";
+                currentElementChildren[1].style.display = "none";
+                currentElement.classList.remove("clipboard-btn--success");
+            }, AppValues.LONG_DELAY_MS);
+            currentElementChildren[0].style.display = "none";
+            currentElementChildren[1].style.display = "inline-block";
+            currentElement.classList.add("clipboard-btn--success");
```
zaychenko-sergei, line 86: Looks very low level and unclear.
Maybe naming the element's role and children roles may help to understand this.

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    private extractSqlQueryFromRoute(): MaybeNull<string> {
+        const sqlQueryFromRoute: MaybeUndefined<string> = this.activatedRoute.snapshot.queryParams[
+            ProjectLinks.URL_QUERY_PARAM_SQL_QUERY
+        ] as MaybeUndefined<string>;
+        return sqlQueryFromRoute ?? "";
+    }
```
zaychenko-sergei, line 95: How it can be null, if you have a default value?

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    public routeToDataset(alias: string): string {
+        return alias.includes("/") ? alias : `kamu/${alias}`;
```
zaychenko-sergei, line 98: Don't understand this. We have other users than `kamu`, what is the hack doing?

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+            case "SqlDataFusion":
+                return "Sql Data Fusion";
+            case "JsonAoA":
+                return "Json AoA";
+            case "ArrowJson":
+                return "Arrow Json";
+            default:
+                return "Unknown options";
+        }
+    }
```
zaychenko-sergei, line 116: A lot of doubt about spelling.
Sql => SQL
Data Fusion => DataFusion
Json => JSON

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    private columnNames(): string[] {
+        return this.sqlQueryExplainer?.output?.schema.fields.map((item) => item.name) as string[];
+    }
```
zaychenko-sergei, line 120: Make it accept a parameter that is already checked against null/undefined.
I.e. `this.sqlQueryExplainer.output.` could come from HTML template within checked divs.

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    public get tableSource(): DataRow[] {
+        const result = this.sqlQueryExplainer?.output?.data.map((item) => {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+            return Object.assign({}, ...item.map((key, index) => ({ [this.columnNames()[index]]: key })));
+        }) as DataRow[];
+        return result;
+    }
```
zaychenko-sergei, line 128: When I see a type cast pragma, it means the code is wrong.

src/app/components/query-explainer/query-explainer.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    private appConfigService = inject(AppConfigService);
+    private httpClient = inject(HttpClient);
+    private toastrService = inject(ToastrService);
+
+    public proccessQuery(query: string): Observable<QueryExplainerResponse> {
+        const url = new URL(`${this.appConfigService.apiServerHttpUrl}/query`);
+        const body = {
+            query: query,
+            dataFormat: "JsonAoA",
+            schemaFormat: "ArrowJson",
```
zaychenko-sergei, line 21: Don't we have these constants in GraphQL generated client?

src/app/components/query-explainer/query-explainer.types.ts
```diff
@@ -0,0 +1,41 @@
+export interface QueryExplainerResponse {
+    input: {
+        query: string;
+        queryDialect: string;
+        dataFormat: string;
+        schemaFormat: string;
```
zaychenko-sergei, line 6: These look like they are enums by nature

src/app/components/query-explainer/query-explainer.types.ts
```diff
@@ -0,0 +1,41 @@
...
+    };
+    commitment: {
+        inputHash: string;
+        outputHash: string;
+        subQueriesHash: string;
+    };
+    proof: {
+        proofValue: string;
+        type: string;
+        verificationMethod: string;
```
zaychenko-sergei, line 20: These look like enums too

src/app/components/query-explainer/query-explainer.types.ts
```diff
@@ -0,0 +1,41 @@
...
+    };
+    output?: {
+        data: [string[]];
+        dataFormat: string;
```
zaychenko-sergei, line 24: And this definitely is an enum

src/app/components/query-explainer/query-explainer.types.ts
```diff
@@ -0,0 +1,41 @@
...
+export interface VerifyQueryResponse {
+    ok: boolean;
+    error?: {
+        kind: string;
+        message: string;
+        actual_hash?: string;
+        expected_hash?: string;
+        dataset_id?: string;
+        block_hash?: string;
+    };
```
zaychenko-sergei, line 40: As I understand, this is modeling a few types of errors which may nor may not have some of these parameters.
I prefer if we define a separate clean structure per each type of error that is hidden here, and union them in the field. The code that is using this will first test for `kind`, which is shared, and downcast to particular good error structures depending on the kind.

src/app/dataset-block/metadata-block/block.service.ts
```diff
@@ -50,4 +50,12 @@ export class BlockService {
...
         );
     }
+
+    public requestSystemTimeBlockByHash(datasetId: string, blockHash: string): Observable<string> {
+        return this.datasetApi.getSystemTimeBlockByHash(datasetId, blockHash).pipe(
+            map((data) => {
+                return data.datasets.byId?.metadata.chain.blockByHash?.systemTime ?? "";
+            }),
+        );
+    }
```
zaychenko-sergei, line 60: Why time is returned as string? It should return time

src/app/dataset-view/additional-components/data-component/data.component.html
```diff
@@ -120,22 +120,50 @@ <h2 class="box-title align-items-center pb-3 m-0">Query:</h2>
...
                                     <div class="f5 text-bold">New Dataset from Query</div>
-                                    <div class="text-small color-fg-muted text-normal pb-1">
-                                        Creates new derivative dataset using current query as the transformation.
+                                    <div class="text-small color-fg-muted text-normal pb-1 lh-130">
+                                        Creates new derivative dataset using current query as the transformation
                                     </div>
                                 </div>
                             </button>
                         </div>
+                        <ng-container *ngIf="isAdmin">
```
zaychenko-sergei, line 135: Again, where is the requirement written about opening this only for admins

zaychenko-sergei (changes requested):
1) I recommend to split the main observable into several reusable observables and avoiding state variables in the component. Use operators like `sharedReplay` when reusing lower-level observables to avoid re-querying the source for multiple derived observables.

2) The Query Explainer component looks very big to me. Consider splitting it into more manageable smaller parts.

src/app/app-routing.module.ts
dmitriy-borzenko: I will clarify the requirements

src/app/components/query-explainer/query-explainer.component.html
dmitriy-borzenko: Fixed

src/app/components/query-explainer/query-explainer.component.ts
dmitriy-borzenko: Fixed

src/app/components/query-explainer/query-explainer.types.ts
dmitriy-borzenko: Added separate error interface for all kind of errors

src/app/dataset-block/metadata-block/block.service.ts
dmitriy-borzenko: Fixed, changed to `Date ` type

src/app/components/query-explainer/query-explainer.types.ts
dmitriy-borzenko: Changed, set ` dataFormat: keyof typeof DataBatchFormat;`

src/app/components/query-explainer/query-explainer.types.ts
dmitriy-borzenko: No, in this place string 


image

src/app/components/query-explainer/query-explainer.service.ts
dmitriy-borzenko: Yes there are, but only as enum keys

src/app/components/query-explainer/query-explainer.types.ts
dmitriy-borzenko: Fixed

src/app/components/query-explainer/query-explainer.component.ts
dmitriy-borzenko: Fixed, removed pragma

src/app/components/query-explainer/query-explainer.component.ts
dmitriy-borzenko: Fixed

src/app/components/query-explainer/query-explainer.component.ts
dmitriy-borzenko: Fixed

src/app/components/query-explainer/query-explainer.component.ts
dmitriy-borzenko: If we use single-tenant image, we don't receive an account in the alias in the response.

src/app/dataset-view/additional-components/data-component/data.component.html
dmitriy-borzenko: Right, only for admins

src/app/common/app.helpers.ts
```diff
@@ -86,3 +86,18 @@ export function convertSecondsToHumanReadableFormat(seconds: number): string {
...
+        const iconsPair: HTMLCollectionOf<HTMLElement> = htmlButtonElement.children as HTMLCollectionOf<HTMLElement>;
+        setTimeout(() => {
+            iconsPair[0].style.display = "inline-block";
+            iconsPair[1].style.display = "none";
+            htmlButtonElement.classList.remove("clipboard-btn--success");
+        }, AppValues.LONG_DELAY_MS);
+        iconsPair[0].style.display = "none";
+        iconsPair[1].style.display = "inline-block";
+        htmlButtonElement.classList.add("clipboard-btn--success");
+    }
```
zaychenko-sergei, line 102: Add some explaining comments how this works.

src/app/components/query-explainer/query-explainer-routing.module.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const routes: Routes = [
+    {
+        path: "",
+        component: QueryExplainerComponent,
+    },
+];
+
+@NgModule({
+    imports: [RouterModule.forChild(routes)],
```
zaychenko-sergei, line 13: Not sure I understand why we need another routing level, the paths are de-facto empty, there is just 1 component.

src/app/components/query-explainer/query-explainer.service.ts
```diff
@@ -0,0 +1,48 @@
...
+@Injectable({
+    providedIn: "root",
+})
+export class QueryExplainerService {
+    private appConfigService = inject(AppConfigService);
+    private http = inject(HttpClient);
+    private toastrService = inject(ToastrService);
+    private localStorageService = inject(LocalStorageService);
+
+    public proccessQuery(query: string): Observable<QueryExplainerResponse> {
```
zaychenko-sergei, line 18: proccessQuery => processQuery

src/app/components/query-explainer/query-explainer.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class QueryExplainerComponent extends BaseComponent implements OnInit {
+    private clipboard = inject(Clipboard);
+    private queryExplainerService = inject(QueryExplainerService);
+    private blockService = inject(BlockService);
+    public readonly DATE_FORMAT = AppValues.DISPLAY_FLOW_DATE_FORMAT;
+    public readonly VerifyQueryKindError: typeof VerifyQueryKindError = VerifyQueryKindError;
+
+    public sqlQueryExplainerResponse: QueryExplainerResponse;
```
zaychenko-sergei, line 35: I still think we can avoid this variable and use observables only.

src/app/components/query-explainer/query-explainer.component.ts
zaychenko-sergei: Consider discussing with SM a possibility of extending this output view with owner account name:

```
"datasets": [
      {
        "id": "did:odf:fed01b8083fa36c3361cc6bee416e89b09b557ff21f5fa81422295c777fbaaabd7529",
        "alias": "account.tokens.portfolio",
        "blockHash": "f1620daed817dc5d5f889af6a1dbc059e9c1a95ede3cdc11a6cee27ecd44607b0cc9b"
      }
    ],
```

src/app/components/query-explainer/query-explainer.component.html
```diff
@@ -0,0 +1,232 @@
...
+    <app-verify-result-section [verifyResponse]="componentData.sqlQueryVerify" />
+
+    <ng-container *ngIf="componentData.sqlQueryExplainerResponse">
+        <div class="row">
+            <div class="col-md-6">
+                <div class="box p-4 my-2 ms-4 h-100">
+                    <div class="position-relative border border-1 p-2 rounded-3">
+                        <span class="position-absolute text-muted group-box-label text-uppercase">Input data</span>
+                        <div class="mt-4 mx-4">
+                            <div class="fw-semibold">Query:</div>
```
zaychenko-sergei, line 11: I think we can and should extract more components from this.
A proposal:
 - Input data section
 - Commitment data section

These would be very easy to test separately, as they become simply a display panels without much logic.

src/app/dataset-view/additional-components/data-component/data.component.html
```diff
@@ -120,22 +120,50 @@ <h2 class="box-title align-items-center pb-3 m-0">Query:</h2>
...
+                                    (click)="verifyQueryResult()"
+                                    class="select-menu-item flex-items-start"
+                                >
+                                    <div>
+                                        <div class="f5 text-bold">Verify query result</div>
+                                        <div class="text-small color-fg-muted text-normal pb-1 lh-130">
+                                            You can verify the query result and make sure that you can trust it
+                                        </div>
+                                    </div>
+                                </button>
```
zaychenko-sergei, line 149: Would be great to allow opening the explainer page in a separate window.
I assume this would require a link instead of a button.

zaychenko-sergei:
image

1) `limit` keyword not highlighted properly (although works fine in data queries)
2) no connection between `limit` in query and `limit` parameter, similar for `offset` -> `skip` pair

zaychenko-sergei:
image

No display of result, neither a simplest toast message, if `/verify` crashes. (found accidentally with a non-working branch on the backend)

src/app/dataset-view/additional-components/data-component/data.component.ts
```diff
@@ -148,6 +162,51 @@ export class DataComponent extends BaseComponent implements OnInit {
...
+                        switchMap(({ uploadPrepareResponse, bodyObject, uploadHeaders }: UploadPrepareData) =>
+                            this.fileUploadService.uploadFileByMethod(
+                                uploadPrepareResponse.method,
+                                uploadPrepareResponse.uploadUrl,
+                                bodyObject,
+                                uploadHeaders,
+                            ),
+                        ),
+                    );
+                }),
```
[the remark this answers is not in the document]
dmitriy-borzenko: Fixed

src/app/common/app.helpers.ts
dmitriy-borzenko: Added explaining commants

src/app/components/query-explainer/query-explainer-routing.module.ts
dmitriy-borzenko: Yes, you are right, I removed the extra routing.

src/app/components/query-explainer/query-explainer.service.ts
dmitriy-borzenko: Fixed

src/app/components/query-explainer/query-explainer.component.ts
dmitriy-borzenko: Fixed

src/app/components/query-explainer/query-explainer.component.ts
dmitriy-borzenko: Discussed and closed

src/app/dataset-view/additional-components/data-component/data.component.html
dmitriy-borzenko: Done

src/app/components/query-explainer/query-explainer.component.html
dmitriy-borzenko: Created more components

zaychenko-sergei (approved):

Issue #455, opened 2024-10-24
Improved SQL support in readme markdown


sergiimk:
In [some datasets](https://platform.demo.kamu.dev/kamu/io.codex.tokens.olhcv) we started to add SQL query examples into README files:
![image](https://github.com/user-attachments/assets/50426ef1-76a6-4f47-afac-900b405d3fdf)

It is included as such blocks:
<pre>
```sql
select * from blah
```
</pre>
that markdown vie automatically assigns correct css classes to.

This ticket is to:
- Add syntax highlight support for SQL in markdown
- Add copy to clipboard button for SQL blocks
- Add run button that opens "Data" tab in new window and executes the selected query

dmitriy-borzenko:
The first 2 points are easy to do. There are special plugins for it. As for the "Run" button, there are difficulties and for this you need to write complex scripts
![image](https://github.com/user-attachments/assets/c698f38c-3a59-4acd-b9fd-4a6910314fdd)

PR #472, opened 2024-11-18
Kamu UI 455 improved sql support in readme markdown

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/455

Result:
image

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -60,13 +60,15 @@
             ],
             "scripts": [
               "node_modules/prismjs/prism.js",
+              "node_modules/prismjs/components/prism-sql.min.js",
               "node_modules/prismjs/components/prism-csharp.min.js",
               "node_modules/prismjs/components/prism-css.min.js",
               "node_modules/prismjs/plugins/line-numbers/prism-line-numbers.js",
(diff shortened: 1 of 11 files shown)


sergiimk:
This looks so cool! :+1:

angular.json
```diff
@@ -60,13 +60,15 @@
             ],
             "scripts": [
               "node_modules/prismjs/prism.js",
+              "node_modules/prismjs/components/prism-sql.min.js",
               "node_modules/prismjs/components/prism-csharp.min.js",
               "node_modules/prismjs/components/prism-css.min.js",
```
zaychenko-sergei, line 65: Now the question is do we need CSharp? :-)

zaychenko-sergei:
I've added a query and pressed `Copy`:

Screenshot from 2024-11-19 15-38-22

Here is what I have in the clipboard:

```
SELECT FROM community.youtube WHERE offset > 5

Run
Run
```

angular.json
dmitriy-borzenko: Removed, no, it's not necessary

Issue #450, opened 2024-10-13
Terms and Conditions UI


sergiimk:
## Business requirements
- The [Kamu Terms of Service](https://docs.google.com/document/d/10pumiXeSXjoGh5bTaRNzLJRUgKECVq0owM7Ti0iO-CA/edit?tab=t.0) (TOS) should be presented to users when they create their account or sign in for the first time via 3rd party auth provider
- TOS content may be presented in UI, or as a link to a place where TOS is published (e.g. GitHub keeps TOS in their [documentation website](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service))
- There should be a way to easily disable this feature in WebUI, in case we need it for demos

dmitriy-borzenko:
## Development requirements
- Create a prototype layout in figma for TOS 
- Create a new API that will allow you to register the date of acceptance of the [Kamu Terms of Service](https://docs.google.com/document/d/10pumiXeSXjoGh5bTaRNzLJRUgKECVq0owM7Ti0iO-CA/edit?tab=t.0) (TOS)
- Extend runtime configuration with new boolean flag `enableTos`(Web UI)
- Extend schema with new boolean flag `enableTos` 
  (https://github.com/kamu-data/helm-charts/blob/24c52c0974619a26e4e7118526475f6c7c3e73f8/charts/kamu-web-ui/values.schema.json)
- Extend struct WebUIFeatureFlags with new boolean flag `enableTos`
  (https://github.com/kamu-data/kamu-cli/blob/e496fe036b58367a5119a1c25721d8c333ec65af/src/app/cli/src/explore/web_ui_server.rs#L64)

PR #473, opened 2024-11-22
Kamu UI 450 terms and conditions UI

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/450

Result:

image


zaychenko-sergei (approved):
Please fix the changelog, release v0.30.0 is already out

Issue #513, opened 2025-01-06
Modify "Scheduled updates" view


dmitriy-borzenko:
![image](https://github.com/user-attachments/assets/4d22aee2-69cb-4e9f-9be0-ad90deb3152c)

1. Having two "Save" buttons is confusing - user shouldn't have to care about what part of config is saved where - there should be only one button to apply the changes

2. "Save Trigger" currently redirects you to "Flows" tab even when the trigger is disabled - so you are redirected but don't really see anything...

3. When trigger is Disabled - I think we should make "Launch every" and "Cron expression" fields disable too - it's too easy currently to configure cron, hit save, and then realize that you forgot to enable the trigger.
If we implement one "Save" button I would also move "Configuration" section under the "Triggers" - it will look strange side-by-side

4. When showing next Cron tick time we should stress that it's UTC

PR #517, opened 2025-01-07
Kamu UI 513 modify scheduled updates view

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/513

Result:

image

image

### DIFF
--- a/src/app/dataset-view/additional-components/dataset-settings-component/tabs/scheduling/dataset-settings-scheduling-tab.component.ts
+++ b/src/app/dataset-view/additional-components/dataset-settings-component/tabs/scheduling/dataset-settings-scheduling-tab.component.ts
@@ -18,6 +18,7 @@ import {
 } from "./dataset-settings-scheduling-tab.component.types";
 import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
 import { ToastrService } from "ngx-toastr";
+import { EMPTY, switchMap } from "rxjs";
 
 @Component({
     selector: "app-dataset-settings-scheduling-tab",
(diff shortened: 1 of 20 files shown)


src/app/dataset-view/additional-components/dataset-settings-component/tabs/scheduling/dataset-settings-scheduling-tab.component.ts
```diff
@@ -73,22 +62,52 @@ export class DatasetSettingsSchedulingTabComponent extends BaseComponent {
...
+                            triggerInput: this.setPollingTriggerInput(this.pollingForm),
+                            datasetInfo: {
+                                accountName: this.datasetBasics.owner.accountName,
+                                datasetName: this.datasetBasics.name,
+                            },
+                        });
+                    } else {
+                        this.toastrService.error("Configuration is not saved");
+                        return EMPTY;
+                    }
```
zaychenko-sergei, line 96: Where is a similar handling when trigger is not saved?

src/app/dataset-view/additional-components/dataset-settings-component/tabs/scheduling/ingest-trigger-form/ingest-trigger-form.component.ts
```diff
@@ -23,7 +23,7 @@ import { DatasetSchedulingService } from "../../../services/dataset-scheduling.s
 })
 export class IngestTriggerFormComponent extends BaseComponent implements OnInit {
```
zaychenko-sergei, line 24: Note: potentially we have a design issue here. Period/CRON stuff is not limited to Ingest flow, I think we will have other cases, like Soft Compacting and GC very soon. There is a sense to isolate this part in a sub-component, which you would reuse for other flow types in future.

Not insisting on doing this right now, we can have it done later, as soon as we need the 2nd use.

dmitriy-borzenko: We will definitely do it, but later

src/app/dataset-view/additional-components/dataset-settings-component/tabs/scheduling/dataset-settings-scheduling-tab.component.ts
dmitriy-borzenko: Error handling is in the method itself when handling triggers.

Issue #502, opened 2024-12-12
Improve representation of retractions and corrections in data tables


sergiimk:
Kamu uses a [special mechanism](https://docs.kamu.dev/cli/transform/retractions-corrections/#changelog-stream-two-events) to represent retractions and corrections in data.

The `op` column encodes operation type as numbers:

| value |   operation    | code  |
| :---: | :------------: | :---: |
|   0   |    `append`    | `+A`  |
|   1   |   `retract`    | `-R`  |
|   2   | `correct-from` | `-C`  |
|   3   |  `correct-to`  | `+C`  |

In `kamu-cli` we already customize representation of `op` column to display such values as codes:
![image](https://github.com/user-attachments/assets/2cde8d1d-9254-4b2d-932d-98264e6453ce)

We should similarly improve the represetnation of this column in Web UI.

Proposal:
- Look for `op` column in data table and replace numeric values with `code`
- Highlight corrections with light yellow color and retractions with light red

Note that detecting column by "op" name is not a reliable solution. In future we will extend the schema to include special metadata in column schema that can be used to reliably tell that some column represents the "operation type".

PR #521, opened 2025-01-09
Kamu UI 502 improve representation of retractions and corrections in data tables

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/502

Result:

image

### DIFF
--- a/src/app/common/app.helpers.ts
+++ b/src/app/common/app.helpers.ts
@@ -147,3 +147,37 @@ export function addMarkdownRunButton(sqlQueries: RegExpMatchArray | null, path: 
         });
     }
 }
+
+export function operationColumnMapper(value: string | number): string {
+    if (typeof value === "number") {
+        switch (value) {
+            case 0:
+                return "+A";
+            case 1:
+                return "-R";
+            case 2:
+                return "-C";
+            case 3:
+                return "+C";
+            /* istanbul ignore next */
+            default:
+                throw new Error("Unknown operation type");
+        }
+    } else return value;
+}
+
+export function setOperationColumnClass(value: string | number): string {
+    if (typeof value === "number") {
+        switch (value) {
+            case 1:
+                return "retraction";
+            case 2:
+            case 3:
+                return "correction";
+
+            /* istanbul ignore next */
+            default:
+                return "";
+        }
+    } else return "";
+}
--- a/src/app/common/app.helpers.ts
+++ b/src/app/common/app.helpers.ts
@@ -3,7 +3,7 @@ import cronParser from "cron-parser";
 
 import { MaybeNull, MaybeNullOrUndefined } from "./app.types";
 import { DataSchema } from "../api/kamu.graphql.interface";
-import { DatasetSchema, OperationColumnClassEnum } from "../interface/dataset.interface";
+import { DatasetSchema } from "../interface/dataset.interface";
 import AppValues from "./app.values";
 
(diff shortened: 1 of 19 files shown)


src/app/components/dynamic-table/dynamic-table.component.html
```diff
@@ -8,10 +8,11 @@
                 [title]="element[disCol]"
                 mat-cell
                 class="truncate-cell"
+                [ngClass]="setOperationColumnClass(element, disCol)"
                 *matCellDef="let element"
                 [attr.data-test-id]="'column-name-' + element[disCol]"
             >
-                {{ element[disCol] }}
+                {{ disCol === "op" && dataRows?.length ? operationColumnMapper(element[disCol]) : element[disCol] }}
```
zaychenko-sergei, line 15: Dynamic table is a dumb presentation-level component that should not know anything about the meaning of the data it displays. It is actually used in many scenarios unrelated to displaying data in our system, i.e. it displays schemas or environment variable.

Design a better solution where we don't have to hardcode "op" column and choose styling depending on the column name. The knowledge of what "op" means does not belong to the presentation level components. Instead, consider designing a model. where you have a higher level data row, or maybe a smarter schema column model, so that it can be given prepared data and displays it without deciding what it should display.

src/app/common/app.helpers.ts
```diff
@@ -147,3 +147,37 @@ export function addMarkdownRunButton(sqlQueries: RegExpMatchArray | null, path:
...
+            case 2:
+                return "-C";
+            case 3:
+                return "+C";
+            /* istanbul ignore next */
+            default:
+                throw new Error("Unknown operation type");
+        }
+    } else return value;
+}
```
zaychenko-sergei, line 167: Let's have a better place for this. It is not a general-purpose application wide utility helper. It is specific to presenting data.

src/app/components/dynamic-table/dynamic-table.component.html
```diff
@@ -7,7 +7,8 @@
             <td
                 [title]="element[disCol]"
                 mat-cell
-                class="truncate-cell"
+                [ngClass]="disCol === 'op' ? element['cssClass'] : ''"
```
zaychenko-sergei, line 10: Comparison with `op` still stays here. Should not be there.

src/app/components/dynamic-table/dynamic-table.component.ts
```diff
@@ -46,7 +47,12 @@ export class DynamicTableComponent implements OnInit, OnChanges, AfterContentIni
             // Casual case, displaying data
         } else {
             this.displayedColumns = this.schemaFields.map((f: DataSchemaField) => f.name);
-            this.dataSource.data = this.dataRows;
+            const modifiedRows = this.dataRows.map((x: DataRow) => ({
+                ...x,
+                op: operationColumnMapper(x.op),
+                cssClass: setOperationColumnClass(x.op),
+            }));
+            this.dataSource.data = modifiedRows;
```
zaychenko-sergei, line 55: Should not happen in presentation component. Data should come in a form that will not take dynamic table component to think how to present it. It is not it's function to think what is this data.

dmitriy-borzenko: Fixed

src/app/components/dynamic-table/dynamic-table.component.html
dmitriy-borzenko: Fixed

Issue #524, opened 2025-01-09
Save sql request in the URL after completing the request


dmitriy-borzenko:
1. 
![image](https://github.com/user-attachments/assets/adff47e4-8691-4f81-92b7-5cecc74479eb)

2. 
![image](https://github.com/user-attachments/assets/84283ad0-c17b-4da6-a1b8-076e29c570ca)

PR #525, opened 2025-01-09
Kamu UI 524 save sql request in the url after completing the request

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/524

Result:

image

image


zaychenko-sergei:
I think we need a similar solution to preserve the query when we are clicking between tabs.
Currently clicking on another dataset tab, and then clicking back to Data tab, resets the query to default state.

zaychenko-sergei (approved):

PR #537, opened 2025-01-30
Kamu UI 531 restructuring of the application catalogs

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/531


zaychenko-sergei:
1) 
In all `NgModule` declarations and imports, please sort them:
 - external modules/components go first, our modules/components go next
 - in each of these groups, please sort alphabetically

2)
`AppHeaderComponent` - belongs to `AppModule`, lives in `src/app/common/components`
`GithubCallbackComponent` - belongs to `AppModule`, lives in `src/app/auth/github-callback`
`NotificationIndicatorComponent` - belongs to `AppModule`, lives in `src/app/common/components`
`AdminDashboardComponent` - belongs to `AppModule`, lives in `src/app/admin-view`

3) 
`DatasetModule` - lives in `src/app/dataset-view`, should we rename it to `DatasetViewModule`, or maybe `src/app/dataset`?
Similar question for `DatasetComponent`.

`DatasetModule` is too large. Consider splitting it into sub-modules (for tabs, wizards, ..).
We will likely want to split it by lazy routing parts on the next phase.

`OverviewHistorySummaryHeaderComponent` - belongs to `DatasetModule`, lives in src/app/common/components
`PageNotFoundComponent` - belongs to `DatasetModule`, lives in `src/app/common/components` (looks like it should be in `AppModule`?)
`ReturnToCliComponent`  - belongs to `DatasetModule`, lives in `src/app/common/components` (looks like it should be in `AppModule` or `LoginModule`?)

4) 
`ToggleComponent` - belongs to `MetadataBlockModule`, lives in `src/app/common/components`, does not seem to be used by anything.

5) 
`DatasetListItemComponent` - belongs to `DatasetListModule`, lives in `src/app/conmon/components`.
Probably it should live inside `src/app/common/components/dataset-list-component`.

6)
`CustomPaginationComponent` - lives in `src/app/components`, has own module. But is used only in `DatasetModule`
`DatasetVisibilityComponent` - lives in `src/app/components`, has own module. But is used only in `DatasetModule`
`LineageGraphComponent` - lives in `src/app/components`, has own module. But is used only in `DatasetModule`
`TimelineComponent` - lives in `src/app/components`, has own module. But is used only in `DatasetModule`

7)
`SearchSidenavModule` and it's content - seems it is not used anywhere.

8)
A lot of doubts about `src/app/services`. I think only really global things like `ErrorHandlerService` should live there.
Some of those services look like they belong to feature modules. 
Please check the locality of their usage for each service separately, and decide whether it's global or feature-specific.

9)
We have `src/app/common/types` and `src/app/interfaces`.
Double check what is global and leave only global definitions in 1 of these places, eliminate the other one.
If something is feature-specific, move it to the feature folder.

zaychenko-sergei (approved):
Great!

Issue #547, opened 2025-02-07
SetTransform UX: editing the transform selects wrong engine


dmitriy-borzenko:
##  Editing the transform selects wrong engine
Add initial transform to a new dataset using DataFusion engine:

![Image](https://github.com/user-attachments/assets/0845964c-0cea-43e3-b168-b04030a7724b)

Now go to "Metadata" tab and click "Edit Transformation". You will see the same SQL, but the engine will be reset to Spark, causing me to break dataset a few times when I was making small SQL edits.

PR #549, opened 2025-02-07
Kamu UI 547 set transform ux  editing the transform selects wrong engine

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/547

Result:

image

image


zaychenko-sergei (changes requested):
Change is fine, but I don't see a test.

dmitriy-borzenko:
> Change is fine, but I don't see a test.

===> Added unit tests

Issue #504, opened 2024-12-13
Angular 16: Simplify boileplate code to read route and query parameters from `ActivatedRoute`


zaychenko-sergei:
Refactor components in the project to leverage Angular 16 new ability to inject route and query parameters as component inputs. This feature simplifies component design by directly mapping route and query parameters to component `@Input` properties, improving code clarity and reducing boilerplate in routing logic.

Motivation:
- This reduces the need for manual extraction of route/query parameters in components.
- This clearly defines component inputs and makes the relationship between the route/query parameters and the component explicit.
- Direct input injection facilitates more predictable and testable components. Complex mocking of ActivatedRoute can be eliminated, setting inputs in tests is much easier.

TODO in this ticket:
- Locate components that currently retrieve route or query parameters using `ActivatedRoute`
- Replace manual parameter extraction in these components with `@Input` properties.
- Update the routing configuration to pass the required parameters as inputs to the components (`withComponentInputBinding`).
- Update unit tests for refactored components to verify the new input mechanism, get rid of older mocking of `ActivatedRoute`
 
Additional reading:
- [How to Pass Url Query Parameters (Routing Params) to Input() Decorator with Transform in Angular](https://medium.com/havelsan/how-to-pass-url-query-parameters-routing-params-to-input-decorator-with-transform-in-angular-fcbe6391ce20) 
- [How to Use @Input() to Read Angular Route Parameters](https://www.freecodecamp.org/news/use-input-for-angular-route-parameters/)

PR #568, opened 2025-02-25
Kamu UI 504 simplify boileplate code to read route and query parameters from activated route

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/504


zaychenko-sergei (approved):
Looks good. 
1) Please don't merge until my clarification
2) Please report another ticket to work on dataset info resolver - title and very basic example would be enough. We will not address it right now, but I'd like to remember about our findings.
3) We also discussed another issue with history traversal innefficiency for SetPollingSource/AddPushSource/SetTransform editors. Plz report it as well.

Issue #605, opened 2025-03-26
Broken "Settings" tab for GitHub user


dmitriy-borzenko:
Current behaviour:

1. Login as  GitHub user
2. Go to  https://platform.demo.kamu.dev/kamu/account.tokens.portfolio?tab=settings

![Image](https://github.com/user-attachments/assets/02efd38b-c6c4-4b6f-bf70-0699264d84e9)

dmitriy-borzenko:
Not relevant

PR #606, opened 2025-03-26
Kamu UI 605 broken settings tab for GitHub user

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/605

Result:
image


dmitriy-borzenko:
Updated the backend - no longer relevant

Issue #619, opened 2025-04-04
Implementation of semantic search


dmitriy-borzenko:
Add semantic search if regular search did not give results

PR #621, opened 2025-04-07
Kamu UI 619 implementation of semantic search

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/619

### DIFF
--- a/src/app/app-config.model.ts
+++ b/src/app/app-config.model.ts
@@ -37,6 +37,7 @@ export interface FeaturesRuntimeConfig {
 
 export interface AppUIConfig {
     ingestUploadFileLimitMb: number;
+    semanticSearchScore: number;
     featureFlags: AppUIConfigFeatureFlags;
 }
 
--- a/src/app/app-config.model.ts
+++ b/src/app/app-config.model.ts
@@ -37,7 +37,7 @@ export interface FeaturesRuntimeConfig {
 
 export interface AppUIConfig {
     ingestUploadFileLimitMb: number;
-    semanticSearchScore: number;
+    semanticSearchScore?: number;
     featureFlags: AppUIConfigFeatureFlags;
 }
 
--- a/src/app/app-config.model.ts
+++ b/src/app/app-config.model.ts
@@ -37,7 +37,7 @@ export interface FeaturesRuntimeConfig {
 
 export interface AppUIConfig {
     ingestUploadFileLimitMb: number;
-    semanticSearchScore?: number;
+    semanticSearchTresholdScore?: number;
     featureFlags: AppUIConfigFeatureFlags;
 }
 
--- a/src/app/common/components/dataset-list-component/dataset-list.component.ts
+++ b/src/app/common/components/dataset-list-component/dataset-list.component.ts
@@ -17,16 +17,16 @@ import { SearchMode } from "src/app/interface/search.interface";
 })
 export class DatasetListComponent {
     @Input({ required: true }) public dataSource: DatasetSearchOverviewFragment[];
-    @Input({ required: true }) public totalCount = 0;
+    @Input({ required: true }) public totalCount: number;
     @Input({ required: true }) public resultUnitText: string;
-    @Input({ required: true }) public hasResultQuantity?: boolean = false;
-    @Input({ required: true }) public isClickableRow?: boolean = false;
-    @Input({ required: true }) public searchMode: SearchMode = SearchMode.SEARCH;
+    @Input({ required: true }) public isClickableRow: boolean;
+    @Input({ required: true }) public searchMode: SearchMode;
     @Input({ required: true }) public sortOptions: {
         value: string;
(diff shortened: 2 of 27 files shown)


src/app/interface/search.interface.ts
```diff
@@ -13,6 +13,12 @@ export interface DatasetSearchResult {
     totalCount: number;
     pageInfo: PageBasedInfo;
     currentPage: number;
+    searchMode: SearchMode;
+}
+
+export enum SearchMode {
+    SEARCH = "search",
```
zaychenko-sergei, line 20: We can call it "text search"

src/app/app-config.model.ts
```diff
@@ -37,6 +37,7 @@ export interface FeaturesRuntimeConfig {
 
 export interface AppUIConfig {
     ingestUploadFileLimitMb: number;
+    semanticSearchScore?: number;
```
zaychenko-sergei, line 40: It's not simply a score, but a "treshold score".

src/app/common/resolvers/search.resolver.ts
```diff
@@ -9,10 +9,28 @@ import { inject } from "@angular/core";
...
     const page = route.queryParamMap.get(ProjectLinks.URL_QUERY_PARAM_PAGE) ?? 1;
     const query = route.queryParamMap.get(ProjectLinks.URL_QUERY_PARAM_QUERY) ?? "";
-    return searchService.searchDatasets(query, Number(page) - 1);
+
+    return searchService.searchDatasets(query, Number(page) - 1).pipe(
+        switchMap((result: DatasetSearchResult) => {
+            return result.datasets.length ||
+                !query ||
+                !isAuthenticated ||
+                typeof isSemanticSearchAvailable === "undefined"
```
zaychenko-sergei, line 31: Exactly what I'm talking about. Boolean should be "true" or "false". As soon as you add "undefined", await for bugs and weird expressions that could be avoided.

src/app/common/resolvers/search.resolver.ts
```diff
@@ -9,10 +9,28 @@ import { inject } from "@angular/core";
...
+                ? of(result)
+                : searchService.semanticSearchDatasets(query);
+        }),
```
zaychenko-sergei, line 34: The logic is not clear enough. Try to simpify it and maybe add some examples in comment.

"If I have results from default method, or query is empty, or I'm not authenticated, or semantic search is not enabled, I get default method results, otherwise, I'm redirected to to semantic search."

Why authentication is a condition? 
If semantic search really requires it, I'm a bit surprised, but if that is true, can you take this into account when assigning "isSemanticSearchAvailable" initially.

src/app/common/components/dataset-list-component/dataset-list.component.ts
```diff
@@ -20,9 +21,14 @@ export class DatasetListComponent {
     @Input({ required: true }) public resultUnitText: string;
     @Input({ required: true }) public hasResultQuantity?: boolean = false;
     @Input({ required: true }) public isClickableRow?: boolean = false;
```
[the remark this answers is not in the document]
dmitriy-borzenko: Fixed

src/app/interface/search.interface.ts
dmitriy-borzenko: Renamed

src/app/app-config.model.ts
dmitriy-borzenko: Renamed

src/app/common/resolvers/search.resolver.ts
dmitriy-borzenko: Fixed

src/app/common/resolvers/search.resolver.ts
dmitriy-borzenko: Fixed

src/app/common/resolvers/search.resolver.ts
```diff
@@ -9,10 +9,28 @@ import { inject } from "@angular/core";
...
+import { LoggedUserService } from "src/app/auth/logged-user.service";
+import { AppConfigService } from "src/app/app-config.service";
 
-export const searchResolver: ResolveFn<void> = (route: ActivatedRouteSnapshot) => {
+export const searchResolver: ResolveFn<Observable<DatasetSearchResult>> = (route: ActivatedRouteSnapshot) => {
     const searchService = inject(SearchService);
+    const loggedUserService = inject(LoggedUserService);
+    const appConfigService = inject(AppConfigService);
+    const isAuthenticated = loggedUserService.isAuthenticated;
+    const isSemanticSearchAvailable = appConfigService.semanticSearchScore;
```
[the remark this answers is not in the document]
dmitriy-borzenko: Added new explicit condition

zaychenko-sergei (approved):

Issue #674, opened 2025-06-10
Behavior for an invalid token in UI


dmitriy-borzenko:
I had an expired token in the local store and when reopening the site, the "Login" button is not shown, so I'm not logged in and cannot login at the same time. 2nd  reload of the page fixes the problem

PR #675, opened 2025-06-11
Improved behavior for an invalid token

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/674

Result:

image

image


zaychenko-sergei (approved):
Some tests would be nice

Issue #573, opened 2025-03-05
Ineffective filtration for BaseYamlEventService


dmitriy-borzenko:
```
  public getEventAsYaml(
        info: DatasetInfo,
        typename: SupportedEvents,
        sourceName: MaybeNull<string> = null,
    ): Observable<MaybeNullOrUndefined<string>> {
        return this.datasetService
            .getDatasetHistory(info, BaseYamlEventService.HISTORY_PAGE_SIZE, this.currentPage)
            .pipe(
                expand((h: DatasetHistoryUpdate) => {
                    const filteredHistory = this.filterHistoryByType(h.history, typename, sourceName);
                    return filteredHistory.length === 0 && h.pageInfo.hasNextPage
                        ? this.datasetService.getDatasetHistory(
                              info,
                              BaseYamlEventService.HISTORY_PAGE_SIZE,
                              h.pageInfo.currentPage + 1,
                          )
                        : EMPTY;
                }),
                map((h: DatasetHistoryUpdate) => {
                    this.history = h;
                    const filteredHistory = this.filterHistoryByType(h.history, typename, sourceName);
                    return filteredHistory;
                }),
                switchMap((filteredHistory: MetadataBlockFragment[]) =>
                    iif(
                        () => !filteredHistory.length,
                        of(null),
                        zip(
                            this.blockService.metadataBlockAsYamlChanges,
                            this.blockService.requestMetadataBlock(info, filteredHistory[0]?.blockHash),
                        ),
                    ),
                ),
                map((result: MaybeNull<[string, unknown]>) => {
                    if (result) return result[0];
                    else return null;
                }),
                last(),
            );
    }
```

PR #713, opened 2025-07-31
Kamu UI 573 fixed/ineffective filtration for base yaml event service

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/573


zaychenko-sergei (commented):
Changes are fine, hopefully the API will finally settle like this

src/app/dataset-block/metadata-block/block.service.ts
```diff
@@ -40,4 +46,15 @@ export class BlockService {
...
+    public requestBlocksByEventType(params: {
+        accountName: string;
+        datasetName: string;
+        eventTypes: [MetadataEventType];
+        encoding: MetadataManifestFormat;
+    }): Observable<MetadataBlockExtended[]> {
+        return this.datasetApi
+            .getBlocksByEventType(params)
+            .pipe(map((data) => data.datasets.byOwnerAndName?.metadata.metadataProjection as MetadataBlockExtended[]));
+    }
```
zaychenko-sergei, line 59: The API level is universal, but complicated. It should be encapsulated and simplified for UI use cases here at service level boundary. 

You are never looking for anything advanced, you are always looking for 1 type of event only, and you never care about manifest format.

The service should provide much simpler methods, something like:
 - getSetTransformBlock(accountName, datasetName) - return 1 block only
 - getSetPollingSourceBlock(accountName, datasetName) - return 1 block only
 - getAddPushSourceBlocks(accountName, datasetName) - return N blocks potentially.
 
 Btw, let's test push sources better. What happens if we have multiple push sources and had multiple edits of them in the history? I would expect UI to see last blocks per push source only. Does this work well?

Issue #714, opened 2025-07-31
Check status for “Saved queries” in “Data” tab


dmitriy-borzenko:
“Saved queries” in “Data” tab is an unimplemented feature, but it’s not grayed out / disabled like the rest. Clicking on it doesn’t do anything and is confusing.

<img width="734" height="461" alt="Image" src="https://github.com/user-attachments/assets/d5f59f9a-bea2-4524-9c74-bd71a496ef72" />

PR #722, opened 2025-08-01
Kamu UI 714 check status for saved queries in data tab

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/714

Result:
Demo mode
image

Production mode
image


sergiimk (approved):
:+1:

Issue #716, opened 2025-07-31
Search page: modify view


dmitriy-borzenko:
On the “Search” view the icons next to datasets show whether they are “root” or “derivative”. These icons are kinda ugly and no one really knows what they mean.

Suggestion:

- Replace dataset icons with avatars off the dataset owners - this will help people to differentiate datasets from one another visually
- Make avatar clickable - taking you to the owner account page
- Put root/derivative as tags next to public/private

<img width="1399" height="731" alt="Image" src="https://github.com/user-attachments/assets/33a0641d-3e99-41b7-973a-a00148ddd65a" />

PR #723, opened 2025-08-01
Kamu UI 716 search page  modify view

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/716

Result:

image


src/app/common/components/dataset-list-component/dataset-list-item/dataset-list-item.component.html
```diff
@@ -1,16 +1,26 @@
...
             <a
                 [routerLink]="['/', row.owner.accountName, row.name]"
                 attr.data-test-id="dataset-name-button-{{ rowIndex }}"
                 class="dataset-list-container__title mb-1 pl-0"
             >
                 {{ row.owner.accountName }}/{{ row.name }}
             </a>
-            <app-dataset-visibility class="ms-2" [datasetVisibility]="row.visibility"></app-dataset-visibility>
+            <app-dataset-visibility class="ms-2" [datasetVisibility]="row.visibility" />
+            <app-dataset-kind [kind]="row.kind" />
```
sergiimk, line 23: Can we make the color of this badge orange for root datasets and blue for derivative, just like in lineage graph:
image

I'd like to use the same colors throughout the system to create strong visual associations.

src/app/api/gql/fragments/fragment-dataset-search-overview.graphql
```diff
@@ -1,5 +1,21 @@
 fragment DatasetSearchOverview on Dataset {
-  ...DatasetBasics
+  id
+  kind
+  name
+  owner {
+    ...AccountExtended
```
sergiimk, line 6: I was surprised to see that `DatasetBasics` fragment was not enough to display the datasets in search results. I assume this is only because `AccountBasics` (which is part of `DatasetBasics`) does not contain `avatarUrl`.

The separation between `AccountBasics` and `AccountExtended` would make sense if the first one was saving some extra database queries on the server side, but `AccountBasics::accountProvider` already makes the server load the full profile information of the account, so requesting `avatarUrl` adds almost zero work.

I would suggest extending `AccountBasics` with `avatarUrl` instead and changing this fragment back to `DatasetBasics`. We can reserve `AccountExtended` for more detailed stuff in the future.

src/app/common/components/dataset-kind/dataset-kind.component.html
```diff
@@ -0,0 +1,3 @@
+<span data-test-id="dataset-kind" class="label label--secondary v-align-middle badge-background">{{
```
sergiimk, line 1: Can we make the color of this badge orange for root datasets and blue for derivative, just like in lineage graph:
image

I'd like to use the same colors throughout the system to create strong visual associations.

sergiimk (changes requested):

src/app/api/gql/fragments/fragment-dataset-search-overview.graphql
dmitriy-borzenko: I remember that Sergey said that it is better not to just take the `avatarUrl ` if you do not use it (he was talking about productivity). The fragment `DatasetBasics ` is used in many places where the `avatarUrl` is not needed. In my opinion, it is not a very good idea to expand the fragment `DatasetBasics ` with the avatarUrl field.

zaychenko-sergei: `accountProvider` was added recently. If it is called, there is no extra cost for `avatarUrl`.

src/app/common/components/dataset-kind/dataset-kind.component.html
```diff
@@ -0,0 +1,9 @@
+<span
+    [ngClass]="{
+        root: isRoot,
+        derivative: !isRoot,
+    }"
+    data-test-id="dataset-kind"
+    class="label label--secondary v-align-middle"
+    >{{ isRoot ? "Root" : "Derivative" }}</span
+>
```
zaychenko-sergei, line 9: It does not look like this piece has enough responsibilities or reuse potential to really be a separate component. Do we plan to use it anywhere else?

src/app/common/components/dataset-list-component/dataset-list-item/dataset-list-item.component.html
```diff
@@ -1,16 +1,26 @@
 <div class="dataset-list-container py-2">
-    <div class="dataset-list-container__title-row">
-        <div class="dataset-list-container__title-row__dataset-name">
-            <i *ngIf="row.kind === 'ROOT'" class="bi bi-journal-album"></i>
-            <mat-icon *ngIf="row.kind === 'DERIVATIVE'">merge</mat-icon>
+    <div class="dataset-list-container__title-row mb-1">
+        <div class="dataset-list-container__title-row__dataset-name d-flex align-items-center">
```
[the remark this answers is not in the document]
dmitriy-borzenko: Fixed

src/app/api/gql/account/fragments/fragment-account-basics.graphql
```diff
@@ -2,4 +2,5 @@ fragment AccountBasics on Account {
   id
   accountName
   accountProvider
+  avatarUrl
```
zaychenko-sergei, line 5: Should we get rid of AccountExtended? I only see 1 use left in the metadata block fragment.

Issue #717, opened 2025-07-31
Modify transform view block


dmitriy-borzenko:
In the [transform block view](https://platform.demo.kamu.dev/kamu/account.tokens.portfolio/block/f16208aac0e80ae8fec862aa1f203fb68600ad0d1309a0c0dc826d88cd3a9d2c9069a) the DIDs of input datasets are long and wrap.

Suggestion:
 use the same kind of shortened representation for DIDs as on the “Overview” tab with a copy button.

<img width="821" height="210" alt="Image" src="https://github.com/user-attachments/assets/2e742273-0fec-4762-93a6-19ebda91c829" />

PR #724, opened 2025-08-01
Changed display of ID for input section

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/717

Result:

image


sergiimk (approved):
:+1:

PR #784, opened 2025-11-18
Kamu UI 778 implement filters for the datasets tab

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/778

Result:

image


src/app/common/components/dataset-flow-process-card/dataset-flow-process-card.component.ts
```diff
@@ -45,6 +46,7 @@ import { SettingsTabsEnum } from "src/app/dataset-view/additional-components/dat
 export class DatasetFlowProcessCardComponent {
     @Input({ required: true }) public datasetBasics: DatasetBasicsFragment;
     @Input({ required: true }) public summary: FlowProcessSummary;
+    @Input() public isAccountCard = false;
```
zaychenko-sergei: Rename `isAccountCard` into `showFlowsHistoryLink`.
Technically it's the same, but it's not the card's responsibility to think in terms of "accounts", it's supposed to have a dumb switch.

dmitriy-borzenko: fixed

zaychenko-sergei (approved):

PR #790, opened 2025-12-12
Add loading new cards on scroll for account filters

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/789

### DIFF
--- a/src/app/api/dataset-flow.api.ts
+++ b/src/app/api/dataset-flow.api.ts
@@ -84,11 +84,18 @@ export class DatasetFlowApi {
     private datasetFlowsProcessesGQL = inject(DatasetFlowsProcessesGQL);
 
     public getDatasetFlowsProcesses(params: { datasetId: string }): Observable<DatasetFlowsProcessesQuery> {
-        return this.datasetFlowsProcessesGQL.watch(params, noCacheFetchPolicy).valueChanges.pipe(
-            map((result: ApolloQueryResult<DatasetFlowsProcessesQuery>) => {
-                return result.data;
-            }),
-        );
+        return this.datasetFlowsProcessesGQL
+            .watch(params, {
+                ...noCacheFetchPolicy,
+                context: {
+                    skipLoading: true,
+                },
+            })
+            .valueChanges.pipe(
+                map((result: ApolloQueryResult<DatasetFlowsProcessesQuery>) => {
+                    return result.data;
+                }),
+            );
     }
 
     public datasetTriggerIngestFlow(params: {
@@ -321,10 +328,20 @@ export class DatasetFlowApi {
     }
 
     public getDatasetFlowsInitiators(datasetId: string): Observable<DatasetFlowsInitiatorsQuery> {
-        return this.datasetFlowsInitiatorsGQL.watch({ datasetId }, noCacheFetchPolicy).valueChanges.pipe(
-            map((result: ApolloQueryResult<DatasetFlowsInitiatorsQuery>) => {
-                return result.data;
-            }),
-        );
+        return this.datasetFlowsInitiatorsGQL
+            .watch(
+                { datasetId },
+                {
+                    ...noCacheFetchPolicy,
+                    context: {
+                        skipLoading: true,
+                    },
+                },
+            )
+            .valueChanges.pipe(
+                map((result: ApolloQueryResult<DatasetFlowsInitiatorsQuery>) => {
+                    return result.data;
+                }),
+            );
     }
 }


src/app/account/additional-components/account-flows-tab/account-flows-tab.types.ts
```diff
@@ -391,4 +392,10 @@ export interface DashboardFiltersOptions {
     selectedFlowProcessStates: FlowProcessEffectiveState[];
     minConsecutiveFailures: number;
     isFirstInitialization: boolean;
+    applyFilters: boolean;
+}
+
+export interface ResultFlows {
+    totalCount: number;
+    nodes: AccountFlowProcessCard[];
 }
```
zaychenko-sergei: The name is not so good. These are not flows, but cards. And it's more a listing, than a result. I would say `FlowProcessCardListing` or something like that.

src/app/account/additional-components/account-flows-tab/account-flows-subtabs/account-flows-datasets-subtab/account-flows-datasets-subtab.component.ts
```diff
@@ -118,105 +125,87 @@ export class AccountFlowsDatasetsSubtabComponent extends BaseComponent implement
...
-            }),
+        const polling$ = timer(0, environment.delay_polling_ms);
+        const triggerWithInitial$ = this.fetchTrigger$.pipe(startWith(undefined));
+        this.accountFlowsCardsData$ = combineLatest([polling$, triggerWithInitial$]).pipe(
             switchMap(() =>
                 this.accountService.getAccountFlowsAsCards({
                     accountName: this.accountName,
                     page: this.currentPage - 1,
-                    perPage: this.CARDS_FLOW_PROCESSES_PER_PAGE,
+                    perPage: this.processesPerPage,
```
zaychenko-sergei: So, with this algorithm, if we loaded 3 pages, then clicked pause on the last item, we get a refresh from page 1 with 3x size. And then, if i scroll more, I would load 3x next page, instead of normal. Not saying this is clearly wrong, but very suspicious.

src/app/account/additional-components/account-flows-tab/account-flows-subtabs/account-flows-datasets-subtab/account-flows-datasets-subtab.component.ts
```diff
@@ -118,105 +125,87 @@ export class AccountFlowsDatasetsSubtabComponent extends BaseComponent implement
...
+    public triggerFetch(): void {
+        this.fetchTrigger$.next();
+    }
+
+    public onScroll(): void {
+        this.processesPerPage = AppValues.UPLOAD_FLOW_PROCESSES_PER_PAGE;
+        if (this.loadingCards || !this.hasNextPage) {
+            return;
+        }
+        this.currentPage++;
```
zaychenko-sergei: I think you have to recompute "currentPage" more accurately here.

Let's, say I opened the page, started scrolling down, had 3 pages loaded, and clicked "pause" on one of the cards.
My `processesPerPage` would be `3 * UPLOAD_FLOW_PROCESSES_PER_PAGE`, while `currentPage` would be 1.

Then, after pause refresh, I start scrolling down again. I expect to load page #4. You reset back the size of the page to default, but `currentPage` is not recomputed back, it was 1 in 3x page sizes, and now you just add 1 and assume normal page size. I think that would load page #2 instead of page #4.

I recommend to get rid of "page" abstraction in general, and use offsets explicitly. It would be less error-prone.

src/app/account/additional-components/account-flows-tab/account-flows-subtabs/account-flows-datasets-subtab/account-flows-datasets-subtab.component.ts
```diff
@@ -100,15 +103,19 @@ export class AccountFlowsDatasetsSubtabComponent extends BaseComponent implement
...
     private readonly toastrService = inject(ToastrService);
     private readonly accountFlowsFiltersService = inject(AccountFlowsFiltersService);
 
-    public accountFlowsCardsData$: Observable<AccountFlowProcessCardConnectionDataFragment>;
+    public accountFlowsCardsData$: Observable<ResultFlows>;
     public currentPage: number = 1;
+    public hasNextPage = false;
+    public processesCards: AccountFlowProcessCard[] = [];
+    public loadingCards = false;
+    private isToggleCardState = false;
```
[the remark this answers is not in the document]
dmitriy-borzenko: Fixed

Issue #807, opened 2026-01-27
Change the display of schemas for the "Data" tab and "Query" page


dmitriy-borzenko:
1. Remove the "Description" column
2. Add an information icon that will display a description.

PR #808, opened 2026-01-27
Kamu UI 807 change the display of schemas for the data tab and query page

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/807

Result:
image

image

image


src/app/interface/dataset.interface.ts
```diff
@@ -26,16 +27,34 @@ export interface DatasetLineageNode {
...
+    inner?: DataSchemaTypeField;
+    itemType?: DataSchemaTypeField;
+    keyType?: {
+        kind: string;
+    };
+    valueType?: {
+        kind: string;
+    };
+    fields?: DataSchemaField[];
 }
```
zaychenko-sergei: Prefer tagged unions pattern - define interfaces for typical cases, with "kind" field everywhere, so that switch/if statements could cast implicitly:

```
// 1. Define the individual types, each with a unique 'kind' discriminator
interface Circle {
  kind: "circle";
  radius: number;
}

interface Square {
  kind: "square";
  sideLength: number;
}

interface Rectangle {
  kind: "rectangle";
  width: number;
  height: number;
}

// 2. Create the union type
type Shape = Circle | Square | Rectangle;

// 3. Use a switch statement and type narrowing to safely access properties
function getArea(shape: Shape): number {
  switch (shape.kind) {
    case "circle":
      // TypeScript knows 'shape' is a Circle here, with access to 'radius'
      return Math.PI * shape.radius ** 2;
    case "square":
      // TypeScript knows 'shape' is a Square here, with access to 'sideLength'
      return shape.sideLength ** 2;
    case "rectangle":
      // TypeScript knows 'shape' is a Rectangle here, with access to 'width' and 'height'
      return shape.width * shape.height;
    default:
      // Exhaustiveness check: if a new shape is added to the union,
      // the compiler will error here, forcing the developer to handle it
      const _exhaustiveCheck: never = shape; 
      throw new Error(`Unknown shape: ${(_exhaustiveCheck as any).kind}`);
  }
}
```

src/app/common/components/dynamic-table/dynamic-table.interface.ts
```diff
@@ -8,3 +8,35 @@
...
+    Struct = "Struct",
+    Time = "Time",
+    Timestamp = "Timestamp",
+    String = "String",
+}
+
+export enum OdfExtraAttributes {
+    EXTRA_ATTRIBUTE_DESCRIPTION = "opendatafabric.org/description",
+    EXTRA_ATTRIBUTE_TYPE = "opendatafabric.org/type",
+}
```
zaychenko-sergei: It does not feel like ODF definitions belong to low-level table visualization component. This is something logically much higher than just table showing data

src/app/common/components/dynamic-table/dynamic-table.component.ts
```diff
@@ -81,4 +107,38 @@ export class DynamicTableComponent implements OnInit, OnChanges, AfterContentIni
...
+    public showMoreBadge(indexRow: number, indexColumn: number, data: DataSchemaField[]): boolean {
+        return (
+            Boolean(this.hasColumnDescriptions) &&
+            indexColumn === 1 &&
+            !this.hasData &&
+            (Boolean(Object.keys(data[indexRow].extra ?? {}).length) || Boolean(data[indexRow].type?.fields?.length))
+        );
+    }
+
+    public extraDesription(indexRow: number, data: DataSchemaField[]): string {
```
zaychenko-sergei: extraDes**c**ription.

PR #822, opened 2026-02-18
Kamu UI 821 lineage tab  incorrect query on data tab when switching to another dataset

dmitriy-borzenko:
Close: https://github.com/kamu-data/kamu-web-ui/issues/821

image

image


src/app/services/map-query-tracker.service.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export interface DatasetEntry {
+    query: string;
+    timestamp: number;
+}
+
+@Injectable({
+    providedIn: "root",
+})
+export class MapQueryTrackerService {
```
zaychenko-sergei: Does not look like there is any map here

PR #843, opened 2026-05-15
Kamu UI 841 reading the odf schema in sources

dmitriy-borzenko:
Closes: https://github.com/kamu-data/kamu-web-ui/issues/841

Result:
image

image

### DIFF
--- a/src/app/dataset-view/additional-components/metadata-component/components/source-events/add-push-source/add-push-source-form.types.ts
+++ b/src/app/dataset-view/additional-components/metadata-component/components/source-events/add-push-source/add-push-source-form.types.ts
@@ -34,6 +34,7 @@ export interface AddPushSourceEditFormType {
         jsonKind?: ReadKind;
         subPath?: string;
         schema?: string[];
+        ddlSchema?: string[];
         separator?: string;
         encoding?: string;
         quote?: string;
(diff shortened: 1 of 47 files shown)


src/app/dataset-view/additional-components/metadata-component/components/source-events/add-push-source/add-push-source-form.types.ts
```diff
@@ -34,6 +34,7 @@ export interface AddPushSourceEditFormType {
         jsonKind?: ReadKind;
         subPath?: string;
         schema?: string[];
+        ddlSchema?: string[];
```
sergiimk: The `schema?: string[];` probably should be removed. We will convert to `schema` in phase 2, but it will not be a `string[]` type.

src/app/search/mock.data.ts
sergiimk: Should this be `ddlSchema`?

src/app/services/templates-yaml-events.service.ts
```diff
@@ -70,6 +70,16 @@ export class TemplatesYamlEventsService {
             kind: "SetPollingSource",
             ...params,
         };
+
+        const { schema: ddlSchema, ...readRest } = params.read;
```
sergiimk: These seem weird. Should we change the name of `params.read.schema` to `ddlSchema` to avoid this renaming?

src/app/dataset-view/additional-components/metadata-component/components/source-events/add-push-source/add-push-source-form.types.ts
dmitriy-borzenko: removed

src/app/search/mock.data.ts
dmitriy-borzenko: Yeah, fixed

src/app/services/templates-yaml-events.service.ts
dmitriy-borzenko: removed

sergiimk (approved):
Tested locally 👍

